PR #1, opened 2020-11-01
:white_check_mark: fetch_aggregates

antonioandara:
fetch_aggregates will return all keys for a certain account


hoh:
The function `sync_fetch_aggregates` does not seem to exist in aleph-client. Do you think it should be added ?

antonioandara:
> The function `sync_fetch_aggregates` does not seem to exist in aleph-client. Do you think it should be added ?

I updated the client on another pull request to add it

hoh:
Obsolete, replaced by #5

PR #5, opened 2021-04-20
:white_check_mark: added fetch_aggregates

antonioandara:
has same functionality as aggregates.fetch from the Aleph-js library

### DIFF
--- a/src/aleph_client/asynchronous.py
+++ b/src/aleph_client/asynchronous.py
@@ -196,6 +196,18 @@ async def fetch_aggregate(address, key, session=None, api_server=DEFAULT_SERVER)
 sync_fetch_aggregate = wrap_async(fetch_aggregate)
 
 
+async def fetch_aggregates(address, session=None, api_server=DEFAULT_SERVER):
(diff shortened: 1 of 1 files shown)


hoh:
Would it not be simpler to make the `key` argument optional as follows ?

```python
async def fetch_aggregate(address, key=None, session=None, api_server=DEFAULT_SERVER):
    if session is None:
        session = await get_fallback_session()

    query_string = f"?keys={key}" if key else ''

    async with session.get(
        "%s/api/v0/aggregates/%s.json%s" % (api_server, address, query_string)
    ) as resp:
        return (await resp.json()).get("data", dict()).get(key)
```

antonioandara:
I considered it at first but I just wanted to keep some consistency between the python and JS libraries since in the  there's a specific "fetch" function that's different from "fetch_one"

hoh:
Thanks for your feedback. I updated much of the code, adding type hints for example. 

Fixed in 3bfd60de055fe71aa93fb121a73487b3472a053f .

PR #22, opened 2021-07-22
fix mksquashfs to avoid append file instead of overwrite it

aliel:
fix mksquashfs to avoid append file instead of overwrite it if squashfs file exists


hoh:
Thanks !

PR #28, opened 2021-10-10
Add solana support

MHHukiewitz:
GM!
Added solana support with according tests. Tests are modeled after the verification process in pyaleph's https://github.com/aleph-im/pyaleph/blob/master/src/aleph/chains/solana.py. Added pynacl as core dependency and base58 as solana-specific. Please verify that the setup still works for you.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,6 +39,7 @@ install_requires =
     aleph-message>=0.1.13
     eth_account>=0.4.0
     python-magic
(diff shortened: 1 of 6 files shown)


MHHukiewitz:
So, if you don't mind, I added an encrypt() function to BaseAccount, so there is a standardized way to store encrypted messages on aleph for all chains using secp256k1 and curve25519.

PR #33, opened 2021-10-13
Add missing imports to aleph_client

MHHukiewitz:
Just so users can find them easier when importing aleph_client.

fetch_aggregates() was missing entirely from main.py. Is there a reason for this?


hoh:
Thank you for this suggestion. I would like to refactor the API to make asynchronous calls first-class, and this goes in the opposite direction however.

hoh:
I opened this issue to improve this: https://github.com/aleph-im/aleph-client/issues/34

PR #39, opened 2021-10-29
Fix JSON serialization in aggregate creation

lgaroche:
Submit requires JSON serializable dict, the create_aggregate call is missing the change.


hoh:
Thank you !

PR #50, opened 2022-02-10
Fixes typo on startDate and endDate filter params name

geraldinefiser:
Params fixed in:

- get_messages
- get_posts
- watch_messages


hoh:
Thank you !

PR #52, opened 2022-02-15
[Tests] Integration tests for Aleph nodes

odesenfans:
Added integration tests for Aleph Core Channel Nodes to test
the message APIs and the propagation of these messages across
the network. Each test attempts to create a specific object
(post, aggregate) and then checks if the message is visible
on another node in a reasonable amount of time.

Added tests for posts, aggregates and the forget functionality.
Tests for IPFS require a bit more scaffolding and will be added
later on.

Moved the unit tests to the new `tests/unit` repository.


tests/integration/itest_forget.py
```diff
@@ -0,0 +1,122 @@
...
+    # Wait for the message to appear on the receiver. We don't check the values,
+    # they're checked in other integration tests.
+    get_post_response = await wait_matching_posts(
+        create_post_response["item_hash"],
+        lambda response: len(response["posts"]) > 0,
+    )
+    print(get_post_response)
+
+    post_hash = create_post_response["item_hash"]
+    reason = "This well thought-out content offends me!"
```
hoh, line 43: 😅

hoh (approved):
Looks great !

PR #54, opened 2022-02-16
[Aggregates] New command to forget an aggregate

odesenfans:
Added the new `forget-aggregate` command to forget all the messages
making up an aggregate. This command takes the key of the aggregate,
finds all the messages that are part of this aggregate using the API
and then forgets them.


odesenfans:
Depends on #53.

hoh (approved):

PR #55, opened 2022-03-08
Fix: Mypy warned that `file_content` could be None

hoh:


src/aleph_client/asynchronous.py
```diff
@@ -231,7 +231,7 @@ async def create_store(
 
     if magic is None:
         pass
-    elif guess_mime_type is True and "mime_type" not in extra_fields:
+    elif file_content and guess_mime_type is True and ("mime_type" not in extra_fields):
```
odesenfans, line 234: Why `guess_mime_type is True` and not just `guess_mime_type`?

hoh: Fixed

PR #69, opened 2022-07-07
Fix: fix text key encoding

aliel:


src/aleph_client/__main__.py
```diff
@@ -57,7 +57,7 @@ def _load_account(
 
     if private_key_str:
         if not private_key_file:
-            return ETHAccount(private_key_str.encode())
+            return ETHAccount(private_key_str)
```
odesenfans, line 60: Any particular reason for this change? Diving into the code, it seems like you can feed it bytes or strings, it doesn't really matter. I'd just like to understand why you want to change this.

aliel: It was a confusing with another related bug that fixed after this one (https://github.com/aleph-im/aleph-client/pull/67)
but another behavior still there when we provide a private-key as text using the CLI

hoh: OK, thanks !

aliel: regarding the doc => https://eth-account.readthedocs.io/en/stable/eth_account.html#module-eth_account.account
the method accept only key as hex str or bytes not encoded string

src/aleph_client/__main__.py
```diff
@@ -56,10 +56,10 @@ def _load_account(
     """Load private key from a file"""
 
     if private_key_str:
-        if not private_key_file:
-            return ETHAccount(private_key_str.encode())
+        if private_key_str[:2] == '0x':
+            return ETHAccount(bytes.fromhex(private_key_str[2:]))
         else:
-            raise ValueError("Choose between a private key or a file, not both")
+            return ETHAccount(bytes.fromhex(private_key_str))
```
odesenfans, line 62:
```suggestion
if private_key_str.startswith("0x"):
    private_key_str = private_key_str[2:]

return EthAccount(bytes.fromhex(private_key_str))
```

odesenfans (commented):
A small suggestion to remove a bit of duplication.

hoh:
Mypy fails with the error `/opt/aleph-client/src/aleph_client/__main__.py:66: error: Name "EthAccount" is not defined`

PR #74, opened 2022-08-01
Return models instead of dictionaries

hoh:
These changes break the API and make it return models from Aleph-message instead of Python dictionaries.


hoh:
@MHHukiewitz this is work in progress to return Aleph Message models instead of dictionaries in aleph-client.

MHHukiewitz (approved):
LGTM

PR #83, opened 2022-08-15
Update test-build.yml

mrhouzlane:
add brew which works for both and kept the requirement with apt-get only for ubuntu


MHHukiewitz (approved):
Workflow at https://github.com/mrhouzlane/aleph-client/actions/runs/2860136391 LGTM

PR #93, opened 2022-08-18
Cleanup: Typer supports Path, conversion was not required

hoh:


src/aleph_client/__main__.py
```diff
@@ -183,11 +182,11 @@ def upload(
     path_object = Path(path)
```
MHHukiewitz, line 182: Should be removed, if I see it correctly
```suggestion

```

MHHukiewitz (approved):
Otherwise, LGTM

PR #96, opened 2022-08-19
Replacing go-ipfs by Kubo

mrhouzlane:
[Kubo](https://github.com/ipfs/kubo) (go-ipfs) is the earliest and most widely used implementation of IPFS.


mrhouzlane:
Problem : go-ipfs "naming" is starting to be deprecated as ipfs is doing a refactoring rename to kubo. 

Solution: updating with-ipfs.dockerfile to kubo version wich is the new version.

PR #97, opened 2022-08-19
kubo-ipfs

mrhouzlane:
Problem : go-ipfs is going to be deprecated as kubo is the new name with new versions.

Solution : replace go-ipfs by kubo as ipfs distribution.


hoh:
I see that the commit does not follow Aleph's commit style. Can you please update that ?

PR #101, opened 2022-09-06
Tests: OS matrix was not fixed

hoh:
Solution: Specify the versions of macOS and Ubuntu to run tests on.

Also prevent running `brew` on Ubuntu since the reference is `apt`.


odesenfans (approved):
Looks alright to me!

PR #103, opened 2022-09-08
Feature: Could not get a single message easily

hoh:
A user could not easily get a single message from the API server, with a clear error message when failing.

Solution: Add function `get_message` singular.

### DIFF
--- a/src/aleph_client/asynchronous.py
+++ b/src/aleph_client/asynchronous.py
@@ -9,6 +9,7 @@ import threading
 import time
 from datetime import datetime
 from functools import lru_cache
+from typing import Type
 
 from aleph_message.models import (
     ForgetContent,
@@ -27,7 +28,9 @@ from aleph_message.models import (
     MessagesResponse,
 )
 
-from aleph_client.types import Account, StorageEnum
+from aleph_client.types import Account, StorageEnum, GenericMessage
+from .exceptions import MessageNotFoundError, MultipleMessagesError
+from .utils import get_message_type_value
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


src/aleph_client/asynchronous.py
```diff
@@ -610,6 +612,33 @@ async def get_messages(
...
+
+async def get_message(
+    item_hash: str,
+    message_type: Type[GenericMessage] = AlephMessage,
+    channel: Optional[str] = None,
+    session: Optional[ClientSession] = None,
+    api_server: str = settings.API_HOST,
+) -> GenericMessage:
+    """Get a single message from its `item_hash`."""
+    raw_program_message = await get_messages(
```
odesenfans, line 626: Why is it named `raw_program_message`?

src/aleph_client/asynchronous.py
```diff
@@ -610,6 +612,33 @@ async def get_messages(
...
+        hashes=[item_hash],
+        session=session,
+        channels=[channel] if channel else None,
+        api_server=api_server,
+    )
+    if len(raw_program_message.messages) < 1:
+        raise MessageNotFoundError(f"No such hash {item_hash}")
+    assert (
```
odesenfans, line 634: `assert` can be deactivated, I see your point but a proper exception would be better IMO.

```python
messages = raw_program_message.messages
if not messages:
  raise MessageNotFoundError(f"No such hash {item_hash}")
elif len(messages > 2):
  raise TooManyResults("API returned more than one message, which is unexpected.")
```

src/aleph_client/asynchronous.py
```diff
@@ -610,6 +612,33 @@ async def get_messages(
...
+        len(raw_program_message.messages) == 1
+    ), "Maximum one message should be returned"
+    message: GenericMessage = raw_program_message.messages[0]
+    assert isinstance(message, message_type)
```
odesenfans, line 638: Same here, not a big fan of asserts.

odesenfans (commented):
A few comments.

src/aleph_client/asynchronous.py
hoh: Changed.

PR #106, opened 2022-09-12
Fix: Received item_hash is the same as in Aleph Explorer

MHHukiewitz:
Problem: Python client returns Messages with different item_hash from what is seen on Aleph Explorer:

Solution: Remove double calculation of item_hash introduced in PR #74


hoh:
Thanks for reporting the issue. I will also create an issue on `aleph-message` regarding the different it computed hashes.

PR #109, opened 2022-09-23
Fix: Any invalid message in a response crashed the entire response

hoh:
The code serving the API may return messages that are not valid according the latest version of the schemas. Ignore them with an optional log instead of failing the parsing of the entire response.

This is an alternative to https://github.com/aleph-im/aleph-message/pull/27 . Putting this logic in the client library allows users to add more options on how invalid messages should be handled.

### DIFF
--- a/src/aleph_client/asynchronous.py
+++ b/src/aleph_client/asynchronous.py
@@ -24,11 +24,12 @@ from aleph_message.models import (
     AggregateMessage,
     StoreMessage,
     ProgramMessage,
-    MessagesResponse,
 )
+from pydantic import ValidationError
 
 from aleph_client.types import Account, StorageEnum, GenericMessage
 from .exceptions import MessageNotFoundError, MultipleMessagesError
+from .models import MessagesResponse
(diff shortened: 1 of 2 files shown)


src/aleph_client/asynchronous.py
```diff
@@ -572,6 +573,8 @@ async def get_messages(
     end_date: Optional[Union[datetime, float]] = None,
     session: Optional[ClientSession] = None,
     api_server: str = settings.API_HOST,
+    ignore_invalid_messages: bool = True,
+    invalid_messages_log_level: Optional[int] = None,
```
odesenfans, line 577: There's 'logging.NOT_SET' which is equal to 0. Using this would avoid the need for 'if invalid_messages_log_level' checks.

hoh: Using the following then ?
```suggestion
    invalid_messages_log_level: int = logging.NOT_SET,
```

odesenfans (approved):

Issue #117, opened 2022-11-17
Broken aiohttp install on python 3.10 when installing dependencies


AmozPay:
Running into [this](https://github.com/aio-libs/aiohttp/issues/6898) very issue when installing `aleph-client` with `setup.py` on Python 3.10

Solution: upgrade requirement from `aiohttp>=3.8.0` to `aiohttp>=3.8.1`

"The most stable version is (and might be for a while) 3.8.1. I recommend you switch to it, as it will most likely have all the functionality you need, newest bugfixes and work out of the box." - See [here](https://github.com/aio-libs/aiohttp/issues/6898#issuecomment-1231532697)

MHHukiewitz:
Fixed with #121

PR #121, opened 2022-11-28
fixed aiohttp dependency [issue #117]

muddi900:
Fix: aiohttp version was incompatible with Python 3.10

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ setup_requires = pyscaffold>=3.2a0,<3.3a0
 # Add here dependencies of your project (semicolon/line-separated), e.g.
 install_requires = 
     coincurve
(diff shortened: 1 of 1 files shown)


hoh:
Thank you !

hoh:
Why not bump directly to [3.8.3](https://docs.aiohttp.org/en/stable/changes.html#id1) ?
3.8.2 fixes bugs and regressions (but introduces a bug fixed by 3.8.3.

Can you put the reason of the commit in the commit itself in the form `Fix: aiohttp version was incompatible with Python 3.10`.
See your [commit style](https://pad.okeso.net/s/vWFVYWEuo)

muddi900:
added 3.8.3 to the cfg.

Issue #119, opened 2022-11-23
Store the private key in the user's home instead of the current directory


hoh:
The creation of a new private key and file to store it in the current directory is confusing to users and can create issues if disclosed.

The private key should instead:
- [x] be created using a dedicated command, asking for the path, similar to how `ssh-keygen` works.
- [x] the generated key should be stored in:
  -  `$ALEPH_IM_HOME/private-keys/ethereum.key` if `$ALEPH_IM_HOME` is defined
  - `$XDG_DATA_HOME/.aleph-im/private-keys/ethereum.key` if `$XDG_DATA_HOME` is defined
  - `$HOME//.aleph-im/private-keys/ethereum.key` else
- [x] a symlink to the `ethereum.key` file should be created with the name `default.key`

PR #123, opened 2022-12-09
Store the private key in the user's home instead of the current directory

aliel:
issue #119

### DIFF
--- a/src/aleph_client/chains/common.py
+++ b/src/aleph_client/chains/common.py
@@ -77,8 +77,10 @@ def get_fallback_private_key() -> bytes:
             private_key = prvfile.read()
     except OSError:
         private_key = generate_key()
+        os.makedirs(os.path.dirname(settings.PRIVATE_KEY_FILE), exist_ok=True)
         with open(settings.PRIVATE_KEY_FILE, "wb") as prvfile:
             prvfile.write(private_key)
+            os.symlink(settings.PRIVATE_KEY_FILE, os.path.join(os.path.dirname(settings.PRIVATE_KEY_FILE), "default.key"))
 
     return private_key
 
@@ -86,5 +88,6 @@ def get_fallback_private_key() -> bytes:
 def delete_private_key_file():
     try:
         os.remove(settings.PRIVATE_KEY_FILE)
+        os.unlink(os.path.join(os.path.dirname(settings.PRIVATE_KEY_FILE), "default.key"))
     except FileNotFoundError:
         pass
--- a/src/aleph_client/chains/common.py
+++ b/src/aleph_client/chains/common.py
@@ -1,6 +1,7 @@
 import os
 from abc import abstractmethod, ABC
-from typing import Dict
+from pathlib import Path
+from typing import Dict, Optional
 
 from coincurve.keys import PrivateKey
 from ecies import decrypt, encrypt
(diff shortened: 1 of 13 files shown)


src/aleph_client/chains/common.py
```diff
@@ -77,14 +77,17 @@ def get_fallback_private_key() -> bytes:
             private_key = prvfile.read()
     except OSError:
         private_key = generate_key()
+        os.makedirs(os.path.dirname(settings.PRIVATE_KEY_FILE), exist_ok=True)
```
hoh, line 80: Would it not be more consistent to use `pathlib.Path` everywhere ?

tests/unit/test_asynchronous.py
```diff
@@ -185,7 +185,7 @@ async def test_forget():
     _get_fallback_session.cache_clear()
 
     if os.path.exists(settings.PRIVATE_KEY_FILE):
-        os.remove(settings.PRIVATE_KEY_FILE)
+        delete_private_key_file()
```
hoh, line 188: What happens if I run pytest on my PC ? Would I then lose my private key ?
I know this PR does not create that issue, but we should solve it.

src/aleph_client/conf.py
```diff
@@ -38,6 +42,21 @@ class Config:
         case_sensitive = False
         env_file = ".env"
 
-
 # Settings singleton
 settings = Settings()
+
+if settings.CONFIG_HOME is None:
```
hoh, line 48: Should this not run when the settings are initialised instead of in the main flow of the module ? 
Maybe something like https://github.com/aleph-im/aleph-vm/blob/main/vm_supervisor/conf.py#L159 ?

src/aleph_client/conf.py
```diff
@@ -38,6 +42,21 @@ class Config:
...
+    else:
+        home = os.path.expanduser("~")
+        os.environ["ALEPH_CONFIG_HOME"] = os.path.join(home, ".aleph-im")
+
+    settings = Settings()
+
+if str(settings.PRIVATE_KEY_FILE) == "ethereum.key":
+    settings.PRIVATE_KEY_FILE = os.path.join(settings.CONFIG_HOME, "private-keys", "ethereum.key")
+
+if "pytest" in sys.modules:
```
hoh, line 61: This seems a bit magic and dangerous

tests/unit/test_asynchronous.py
aliel: in pytest the private key name now has a suffix "test"

src/aleph_client/commands/account.py
```diff
@@ -0,0 +1,50 @@
...
+    private_key_file = typer.prompt("Enter file in which to save the key", settings.PRIVATE_KEY_FILE)
+
+    if os.path.exists(private_key_file):
+        typer.echo(f"Error: key already exists: '{private_key_file}'")
+        exit(1)
+
+    private_key = None
+    if from_private_key is not None:
+        account: AccountFromPrivateKey = _load_account(from_private_key, None)
+        private_key = account.private_key
```
MHHukiewitz, line 39: account.private_key does not exist as a property

src/aleph_client/conf.py
```diff
@@ -3,16 +3,20 @@
...
 
 
 class Settings(BaseSettings):
+    CONFIG_HOME: Optional[str] = None
+
     # In case the user does not want to bother with handling private keys himself,
     # do an ugly and insecure write and read from disk to this file.
     PRIVATE_KEY_FILE: Path = Field(
-        default=Path("device.key"),
+        default=Path("ethereum.key"),
```
MHHukiewitz, line 16: can also be a Solana key though. I often wondered how we should solve this problem. Maybe different keys for different accounts the user creates?

src/aleph_client/conf.py
MHHukiewitz: Yes, it will immediately run when you import it. Therefore settings.CONFIG_HOME will always be None

MHHukiewitz (commented):
I think we need to rework account and key management quite a bit. This is the right beginning but we need to solve issues like:
- All account types (Ethereum, Solana...) use the same keyfile. Should they be made from the same secret? Are they compatible?
- The current configuration management is lackluster IMHO and offers no possibility of calculating defaults as is done here. We discussed a Client object to handle all this in an OO way.

src/aleph_client/conf.py
hoh: We support Ethereum by default. This is the default setting.

src/aleph_client/conf.py
hoh: Unless the environment variable `ALEPH_CONFIG_HOME` specifies otherwise, since Pydantic will load the environment variables on initialisation.

hoh:
Here is a PR on this branch to never delete private keys. It uses temporary files instead for tests.
https://github.com/aleph-im/aleph-client/pull/136

hoh:
@MHHukiewitz Please do not create merge commits or push on someone else's branch. Use rebase instead. 
For change proposals on an existing pull request, please create your own branch, and either ask the author to rebase/cherry-pick your commits or open a pull request with that branch as target instead of master.

MHHukiewitz:
@hoh I discussed it with Alie prior, but I will respect this workflow next time.

PR #130, opened 2022-12-27
Fix: item type is not set for non-inline messages

odesenfans:
Problem: submitting non-inline messages fails because the item type field is not set.

Solution: set the item type field for ipfs and storage item types.


MHHukiewitz (approved):
LGTM

PR #131, opened 2022-12-27
Fix: allow user content to contain datetimes

odesenfans:
Problem: using datetime objects in message contents (ex: in POST messages) fails because the JSON serializer used in aleph-client does not support datetimes by default.

Solution: add a JSON serializer function for datetimes.


MHHukiewitz:
@odesenfans seems ready to me, no?

MHHukiewitz:
Superseded by https://github.com/aleph-im/aleph-sdk-python/pull/82

Issue #84, opened 2022-08-15
Optional parameters are not immutable to being passed 'None'


MHHukiewitz:
Take the following function signature as an example:
```python
create_post(
    account: Account,
    post_content,
    post_type: str,
    ref: Optional[str] = None,
    address: Optional[str] = settings.ADDRESS_TO_USE,
    channel: str = settings.DEFAULT_CHANNEL,
    session: Optional[ClientSession] = None,
    api_server: str = settings.API_HOST,
    inline: bool = True,
    storage_engine: StorageEnum = StorageEnum.storage,
) -> PostMessage:
```
The problem lies primarily with `address` & `channel`: One of them is not `Optional[str]`, where both provide default values. Another problem is that if I build another function on top like:
```python
async def post_or_amend_object(obj: T, account=None, channel=None):  # note, account and channel would be 'Optional'
    resp = await client.create_post(account, obj.content, channel=channel, ref=obj.item_hash)
```
Then the default values are being overwritten by `None`, so I had to to the following:
```python
    if account is None:
        account = FALLBACK_ACCOUNT
    if channel is None:
        channel = TEST_CHANNEL
    resp = await client.create_post(account, obj.content, channel=channel, ref=obj.item_hash)
```
which doesn't seem to serve the purpose of an `Optional` parameter. The same would be true for the `api_server` and `storage_engine` parameters. This does not seem to be an issue with the `get_posts()` and other `get_` methods. But the same is true for `create_program()` and `create_aggregate()` functions. IMHO, all of the functions of aleph-client should behave similarly when confronted with optional parameters, but because of this issue, this is not the case.

hoh:
This is indeed an issue and we should try to fix it for the next release.

MHHukiewitz:
Fixed in #132

PR #132, opened 2023-01-02
Problem: Functions are not immutable to being passed `None` (see #84)

MHHukiewitz:
Solution: Set `None` as the default value everywhere and handle wanted defaults inside the function. Add docstrings to elucidate default behaviors and parameters.

### DIFF
--- a/src/aleph_client/asynchronous.py
+++ b/src/aleph_client/asynchronous.py
@@ -64,12 +64,11 @@ def get_fallback_session() -> ClientSession:
 
 
 async def ipfs_push(
-        content,
-        session: Optional[ClientSession] = None,
-        api_server: str = settings.API_HOST,
+        content: Dict,
+        session: ClientSession,
+        api_server: str,
 ) -> str:
-    session = session or get_fallback_session()
-
+    """Push arbitrary content as JSON to the IPFS service."""
     url = f"{api_server}/api/v0/ipfs/add_json"
     logger.debug(f"Pushing to IPFS on {url}")
 
@@ -79,12 +78,11 @@ async def ipfs_push(
 
 
 async def storage_push(
-        content,
-        session: Optional[ClientSession] = None,
-        api_server: str = settings.API_HOST,
+        content: Dict,
+        session: ClientSession,
+        api_server: str,
 ) -> str:
-    session = session or get_fallback_session()
-
+    """Push arbitrary content as JSON to the storage service."""
     url = f"{api_server}/api/v0/storage/add_json"
     logger.debug(f"Pushing to storage on {url}")
 
@@ -95,11 +93,10 @@ async def storage_push(
 
 async def ipfs_push_file(
         file_content,
-        session: Optional[ClientSession] = None,
-        api_server: str = settings.API_HOST,
+        session: ClientSession,
+        api_server: str,
 ) -> str:
-    session = session or get_fallback_session()
-
+    """Push a file to the IPFS service."""
     data = aiohttp.FormData()
     data.add_field("file", file_content)
 
@@ -113,11 +110,10 @@ async def ipfs_push_file(
 
 async def storage_push_file(
         file_content,
-        session: Optional[ClientSession] = None,
-        api_server: str = settings.API_HOST,
+        session: ClientSession,
+        api_server: str,
 ) -> str:
-    session = session or get_fallback_session()
-
+    """Push a file to the storage service."""
     data = aiohttp.FormData()
     data.add_field("file", file_content)
 
@@ -131,12 +127,10 @@ async def storage_push_file(
 
 async def broadcast(
         message,
-        session: Optional[ClientSession] = None,
-        api_server: str = settings.API_HOST,
+        session: ClientSession,
+        api_server: str,
 ) -> None:
     """Broadcast a message on the Aleph network via pubsub for nodes to pick it up."""
-    session = session or get_fallback_session()
-
     url = f"{api_server}/api/v0/ipfs/pubsub/pub"
     logger.debug(f"Posting message on {url}")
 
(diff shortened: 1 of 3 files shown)


src/aleph_client/asynchronous.py
```diff
@@ -234,23 +254,49 @@ async def create_aggregate(
 
 async def create_store(
         account: Account,
-        address=settings.ADDRESS_TO_USE,
+        address: Optional[str] = None,
         file_content: Optional[bytes] = None,
+        file_path: Optional[str] = None,
         file_hash: Optional[str] = None,
-        guess_mime_type: bool = False,
+        guess_mime_type: Optional[bool] = None,
```
odesenfans, line 261: IMO that's pushing the logic a bit far to have this as an optional. Any value that doesn't have a default in `settings` does not need to be optional.

src/aleph_client/asynchronous.py
```diff
@@ -234,23 +254,49 @@ async def create_aggregate(
...
         ref: Optional[str] = None,
-        storage_engine=StorageEnum.storage,
+        storage_engine: Optional[StorageEnum] = None,
```
odesenfans, line 263: Same here, it makes sense to use the `storage` engine by default.

src/aleph_client/asynchronous.py
```diff
@@ -301,25 +347,52 @@ async def create_program(
...
-        memory: int = settings.DEFAULT_VM_MEMORY,
-        vcpus: int = settings.DEFAULT_VM_VCPUS,
-        timeout_seconds: float = settings.DEFAULT_VM_TIMEOUT,
-        persistent: bool = False,
-        encoding: Encoding = Encoding.zip,
+        api_server: Optional[str] = None,
+        memory: Optional[int] = None,
+        vcpus: Optional[int] = None,
+        timeout_seconds: Optional[float] = None,
+        persistent: Optional[bool] = None,
```
odesenfans, line 358: Revert.

src/aleph_client/asynchronous.py
```diff
@@ -392,13 +465,28 @@ async def forget(
         account: Account,
         hashes: List[str],
         reason: Optional[str],
-        storage_engine: StorageEnum = StorageEnum.storage,
-        channel: str = settings.DEFAULT_CHANNEL,
-        address: Optional[str] = settings.ADDRESS_TO_USE,
+        storage_engine: Optional[StorageEnum] = None,
```
odesenfans, line 468: Why does `forget` specify a storage engine? Seems weird. I'm gonna give a go to refactoring the broadcast + submit code today, it's getting annoying.

src/aleph_client/asynchronous.py
```diff
@@ -690,13 +856,26 @@ async def watch_messages(
...
+    :param content_types: Content types to watch
+    :param refs: References to watch
+    :param addresses: Addresses to watch
+    :param tags: Tags to watch
+    :param hashes: Hashes to watch
+    :param channels: Channels to watch
+    :param chains: Chains to watch
+    :param start_date: Start date from when to watch
+    :param end_date: End date until when to watch
+    :param session: Session to use (DEFAULT: get_fallback_session())
```
odesenfans, line 874: IMO you can use lowercase, `DEFAULT` everywhere is a bit aggressive on the eyes :p

src/aleph_client/asynchronous.py
```diff
@@ -64,12 +64,11 @@ def get_fallback_session() -> ClientSession:
 
 
 async def ipfs_push(
-        content,
-        session: Optional[ClientSession] = None,
-        api_server: str = settings.API_HOST,
+        content: Dict,
```
odesenfans, line 67: Can be `Mapping` as you don't modify the content of the dictionary.

src/aleph_client/asynchronous.py
```diff
@@ -79,12 +78,11 @@ async def ipfs_push(
 
 
 async def storage_push(
-        content,
-        session: Optional[ClientSession] = None,
-        api_server: str = settings.API_HOST,
+        content: Dict,
```
odesenfans, line 81: Same.

src/aleph_client/asynchronous.py
MHHukiewitz: The problem with these default values is their non-transitiveness, as explained in #84. Am I getting something wrong doing it like this?

src/aleph_client/asynchronous.py
MHHukiewitz: All these `None` are in place to comply better with mypy. The actual default is being set in the function body.

IMO the ideal behavior should be that if "nothing"/`None` is being passed to some argument with a default, then the default should instead be used, but Python overrides the default then with `None`.

src/aleph_client/asynchronous.py
MHHukiewitz: What exactly?

MHHukiewitz:
Any idea why the tests fail now? The absolute import of the `tests` directory doesn't seem to work anymore

src/aleph_client/asynchronous.py
odesenfans: I'm not sure I get what you mean by (non-)transitivity. I get your issue with using the defaults from the `settings` module. But here you have a boolean value which does not depend on a setting, has a sensible default value (false here) and can be overridden by the user. I don't see why changing it to an optional improves the situation.

src/aleph_client/asynchronous.py
odesenfans: Same discussion as above, but `persistent: bool = False` is fine.

src/aleph_client/asynchronous.py
MHHukiewitz: Hm, yeah, setting a boolean yourself is less complex than referencing to the settings, you're right

MHHukiewitz: But still, my issue with this "non-transitiveness" is the following: If at any point of functions calling functions with the same kwargs I want to change the default, I'll have to do it at the beginning of the chain, not at the end, where the value is actually required.

My issue lies with the fact that users, using the SDK, might need to write their own default handling as pointed out in the issue. This is also the case with the booleans, which creates unnecessary overhead for the user. If the functions do not care whether they are passed `None` or not and still use their respective default, then this is ideal behavior.

MHHukiewitz: It is indeed the most stupid behavior that `None` overwrites defaults.

MHHukiewitz: The cases you commented on here are the ones that do not have this chaining of function calls. But if a user decides to write their own module with the SDK and writes a function that passes its arguments to this function, then the developer will have the same headache as I had before.

It is a really weird situation that feels a lot like the `null` and `undefined` double-trouble with Javascript.

odesenfans: My opinion here is that "inner" functions should never try to load a setting from a global configuration object. It introduces the whole mess you're talking about. And you fixed that, so that's good. Now, on the topic of nulls. The good practice in Python is to:

1. Use type hints. if you say something is a `bool`, then the user better pass a bool and not `None`.
2. If you want to really enforce the type hints, a good old 

```
if not isinstance(boolean_value, bool):
  raise TypeError("Get rekt")
```

is the way to go. We should not allow `None` where it doesn't make sense, period. As I mentioned earlier, the good defaults are:

```
def f_with_defaults(b: bool, b_with_sensitive_default: bool = False, l: Optional[List] = None, d: Optional[Dict] = None):
  l = l or []
  d = d or {} 
```

odesenfans (approved):

odesenfans (approved):

PR #133, opened 2023-01-03
Feat: Add LocalVmCache for testing purposes

MHHukiewitz:
Problem: Programs that work on Aleph VMs and use VmCache cannot work locally

Solution: Introduce abstract base class for VmCache and implement LocalVmCache to mock VmCache functionality

### DIFF
--- a/src/aleph_client/vm/cache.py
+++ b/src/aleph_client/vm/cache.py
@@ -1,12 +1,21 @@
 import re
 import fnmatch
 import abc
-from typing import Union, Optional, Any, Dict, List
+from typing import Union, Optional, Any, Dict, List, NewType
 
 from aiohttp import ClientSession
 
 from aleph_client.asynchronous import get_fallback_session
 from ..conf import settings
+
+
+CacheKey = NewType("CacheKey", str)
+
+
+def sanitize_cache_key(key: str) -> CacheKey:
+    if not re.match(r"^\w+$", key):
+        raise ValueError("Key may only contain letters, numbers and underscore")
+    return CacheKey(key)
 
 
 class BaseVmCache(abc.ABC):
@@ -50,9 +59,8 @@ class VmCache(BaseVmCache):
         self.api_host = api_host if api_host else settings.API_HOST
 
     async def get(self, key: str) -> Optional[bytes]:
-        if not re.match(r"^\w+$", key):
-            raise ValueError("Key may only contain letters, numbers and underscore")
-        async with self.session.get(f"{self.api_host}/cache/{key}") as resp:
+        sanitized_key = sanitize_cache_key(key)
+        async with self.session.get(f"{self.api_host}/cache/{sanitized_key}") as resp:
             if resp.status == 404:
                 return None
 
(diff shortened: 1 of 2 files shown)


MHHukiewitz:
@hoh I did not include a test for VmCache, as mocking it is more complex than the code that is actually being tested. All the functionality of VmCache invokes well-known modules and functions.

src/aleph_client/vm/cache.py
```diff
@@ -1,56 +1,121 @@
...
+
+    async def keys(self, pattern: str = "*") -> List[str]:
+        if not re.match(r"^[\w?*^\-]+$", pattern):
+            raise ValueError(
+                "Pattern may only contain letters, numbers, underscore, ?, *, ^, -"
+            )
+        all_keys = list(self._cache.keys())
+        if pattern == "*":
+            return all_keys
+        return [key for key in all_keys if re.match(pattern, key)]
```
odesenfans, line 121: Seems like we're mixing both regular expressions and glob expressions here. What's the actual spec?

src/aleph_client/vm/cache.py
```diff
@@ -1,56 +1,121 @@
...
+class VmCache(BaseVmCache):
+    """Virtual Machines can use this cache to store temporary data in memory on the host."""
+
+    cache: Dict[str, bytes]
+    api_host: str
+
+    def __init__(self, session: Optional[ClientSession] = None, api_host: Optional[str] = None):
+        super().__init__(session)
+        self.cache = {}
+        self.api_host = api_host if api_host else settings.API_HOST
```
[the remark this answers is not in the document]
MHHukiewitz: Yup, I didn't look into how to retrieve the local API. Would that indeed be `localhost` then?

src/aleph_client/vm/cache.py
hoh: Keys may only be composed of letters and digits that `re.match(r"^\w+$", key)`.

odesenfans: Yes, what bothers me here is the pattern. I see `if pattern == "*":` and then `re.match(pattern, key)` so there's something to clarify.

MHHukiewitz: Whoops, you're right! These are wildcard and not regular expressions: https://redis.io/docs/stack/search/reference/query_syntax/

MHHukiewitz:
Can we still merge it though? I understand that you want to increase the coverage percentage, but in this case, I really do not see the advantage in testing the handling of HTTP requests to the VM cache.

src/aleph_client/vm/cache.py
```diff
@@ -1,56 +1,121 @@
...
             if resp.status == 404:
                 return None
 
             resp.raise_for_status()
             return await resp.read()
 
-    async def set(self, key: str, value: Union[str, bytes]):
+    async def set(self, key: str, value: Union[str, bytes]) -> Any:
         if not re.match(r"^\w+$", key):
             raise ValueError("Key may only contain letters, numbers and underscore")
```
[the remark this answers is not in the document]
odesenfans: I don't get your point. This is mostly about factorizing code that we use today and reducing duplication, it wouldn't change anything to the behavior of the code.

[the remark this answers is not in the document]
MHHukiewitz: Ahhh, damn, focused more on the return type. I'm sorry, of course you are right :bow:

odesenfans (approved):

Issue #60, opened 2022-04-30
There is no function to download raw data from Aleph


hoh:
Users should not have to write code such as this themselves:

```python
async def download_data_from_aleph(file_hash) -> bytes:
    async with aiohttp.ClientSession(connector=aiohttp.TCPConnector()) as session:
        async with session.get(
            f"https://aleph.cloud/api/v0/storage/raw/{file_hash}"
        ) as response:
            if response.status != 200:
                response_text = await response.text()
                raise HTTPException(status_code=response.status, detail=response_text)

            return await response.read()
```

MHHukiewitz:
Is it intended only for this endpoint or for all of them?

hoh:
I do not understand your question.

The url prefix `/api/v0/storage/raw/{file_hash}` exposes raw data as bytes from a file stored in Aleph builtin storage.

PR #134, opened 2023-01-03
Feat: Get raw storage file (issue #60)

MHHukiewitz:
Problem: There's no function to download storage data by its item hash (see issue #60)

Solution: Add get_store_file()

### DIFF
--- a/src/aleph_client/asynchronous.py
+++ b/src/aleph_client/asynchronous.py
@@ -562,6 +562,36 @@ async def get_posts(
         return await resp.json()
 
 
+async def get_store_file(
+        file_hash: str,
+        storage_engine: Optional[StorageEnum] = None,
+        session: Optional[ClientSession] = None,
+        api_server: Optional[str] = None
+) -> bytes:
+    """
+    Get a file from the storage engine as raw bytes.
+
+    :param file_hash: The hash of the file to retrieve.
+    :param storage_engine: The storage engine to use. (DEFAULT: "storage")
+    :param session: The aiohttp session to use. (DEFAULT: get_fallback_session())
+    :param api_server: The API server to use. (DEFAULT: "https://api2.aleph.im")
+    """
+    storage_engine = storage_engine or StorageEnum.storage
+    session = session or get_fallback_session()
+    api_server = api_server or settings.API_HOST
+
+    if storage_engine == StorageEnum.storage:
+        async with session.get(f"{api_server}/api/v0/storage/raw/{file_hash}") as response:
+            response.raise_for_status()
+            return await response.read()
+    elif storage_engine == StorageEnum.ipfs:
+        async with session.get(f"{api_server}/api/v0/ipfs/files/read/{file_hash}") as response:
+            response.raise_for_status()
+            return await response.read()
+    else:
+        raise ValueError(f"Invalid storage engine: {storage_engine}")
+
+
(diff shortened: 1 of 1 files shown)


src/aleph_client/asynchronous.py
```diff
@@ -562,6 +562,36 @@ async def get_posts(
...
+    :param storage_engine: The storage engine to use. (DEFAULT: "storage")
+    :param session: The aiohttp session to use. (DEFAULT: get_fallback_session())
+    :param api_server: The API server to use. (DEFAULT: "https://api2.aleph.im")
+    """
+    storage_engine = storage_engine or StorageEnum.storage
+    session = session or get_fallback_session()
+    api_server = api_server or settings.API_HOST
+
+    if storage_engine == StorageEnum.storage:
+        async with session.get(f"{api_server}/api/v0/storage/raw/{file_hash}") as response:
```
odesenfans, line 584: Independently of the storage engine, you can always use `/api/v0/storage/raw/{file_hash}`. There's no `/api/v0/ipfs/files/read/{file_hash}` endpoint afaik.

src/aleph_client/asynchronous.py
```diff
@@ -562,6 +562,36 @@ async def get_posts(
         return await resp.json()
 
 
+async def get_store_file(
```
odesenfans, line 565: `download_file` ?

src/aleph_client/asynchronous.py
```diff
@@ -562,6 +562,36 @@ async def get_posts(
...
+        file_hash: str,
+        storage_engine: Optional[StorageEnum] = None,
+        session: Optional[ClientSession] = None,
+        api_server: Optional[str] = None
+) -> bytes:
```
odesenfans, line 570: Beware of large files, this function might end up using a lot of memory.

MHHukiewitz: Recommend some type of generator then? Added a warning for starters ^^'

odesenfans: A warning is a good start. A bytes generator might be a good idea, as long as what the user is doing with it does not take too long. And then adding higher-level functions on top of the generator, like `def download_file(file_hash: str, output_file: pathlib.Path)` etc.

MHHukiewitz: Judging from https://github.com/aleph-im/pyaleph/blob/190ebef84b5823c4ea5616f5152b4f6fa21b17b9/src/aleph/web/controllers/storage.py#L79
there is currently no way to request a StreamingResponse, so we inadvertently will have the whole file in memory until we store/dispose of it in a file, correct?

hoh: Indeed, but that is on the server side with lots of RAM.

odesenfans (approved):

src/aleph_client/asynchronous.py
hoh: Returning a file descriptor may be ideal. 

Headers contain the size of the content, which could be stored in a temporary file when larger than a certain threshold.

hoh: See https://docs.python.org/3/library/io.html#io.BufferedReader

MHHukiewitz: Let's create an issue for this then :+1:

PR #135, opened 2023-01-09
Feature: support the new POST /messages/ endpoint

odesenfans:
Problem: the `broadcast` function uses the now deprecated /ipfs/pubsub/pub/ endpoint.

Solution: use the new POST /messages/ endpoint. If the selected API server does not support the new endpoint, fall back to the previous implementation.

Used the occasion to clean up the broadcast error handling.

### DIFF
--- a/src/aleph_client/asynchronous.py
+++ b/src/aleph_client/asynchronous.py
@@ -9,7 +9,7 @@ import threading
 import time
 from datetime import datetime
 from functools import lru_cache
-from typing import Type
+from typing import Type, Mapping
 
 from aleph_message.models import (
     ForgetContent,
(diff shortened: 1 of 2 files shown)


odesenfans:
Note: the code in this PR is retro compatible, or at least it's supposed to be. This can be merged at anytime, there's no dependency to the ongoing work on CCNs.

src/aleph_client/asynchronous.py
```diff
@@ -23,7 +23,8 @@
     AlephMessage,
     AggregateMessage,
     StoreMessage,
-    ProgramMessage, ItemType,
+    ProgramMessage,
+    ItemType,
```
hoh, line 27: Can you separate formatting edits from actual code changes ? Mixing them makes Pull Requests more difficult to process and `git blame` confusing.

src/aleph_client/asynchronous.py
```diff
@@ -129,56 +130,129 @@ async def storage_push_file(
         return (await resp.json()).get("hash")
 
 
+def _log_publication_status(publication_status: Mapping[str, Any]):
```
hoh, line 133: Isn't the fact that a method prefixed with `_log_` raises an exception confusing ?

src/aleph_client/asynchronous.py
```diff
@@ -129,56 +130,129 @@ async def storage_push_file(
...
+    ) as response:
+        await _handle_broadcast_deprecated_response(response)
+
+
+async def _handle_broadcast_response(
+    response: aiohttp.ClientResponse, sync: bool
+) -> None:
+    if response.status not in (200, 202):
+        await _handle_broadcast_error(response)
+    else:
```
hoh, line 209: Can you add a test on this behaviour ?

src/aleph_client/asynchronous.py
```diff
@@ -64,9 +65,9 @@ def get_fallback_session() -> ClientSession:
 
 
 async def ipfs_push(
-        content,
-        session: Optional[ClientSession] = None,
-        api_server: str = settings.API_HOST,
```
MHHukiewitz, line 69: The thing about formatting:
For some reason, PyCharm believes the double indent to be the superior one, probably based on [PEP 8 - Indentation](https://peps.python.org/pep-0008/#indentation)
I found it weird at first, too, but after some consideration, I actually prefer this style. The list of parameters is perfectly identifiable against the regular function body, which I find satisfying to look at.

MHHukiewitz (approved):
Besides the option to discuss indentation rules, LGTM

src/aleph_client/asynchronous.py
odesenfans: So, as discussed, let's use black for all our Python codebases to avoid formatting debates.

src/aleph_client/asynchronous.py
odesenfans: Done.

src/aleph_client/asynchronous.py
odesenfans: It only raises a `ValueError` if the API response has the wrong format. We need better error handling in this repository but until then I'd say that this is fair.

PR #138, opened 2023-01-11
Feature: make mypy usable

odesenfans:
Problem: the codebase is not configured to use mypy effectively.
Solution: add a `mypy.ini` file and fix mypy errors.

The `mypy.ini` file is similar to the one used in pyaleph.


MHHukiewitz (approved):
LGTM

PR #139, opened 2023-01-11
Feature: refactor the broadcast function

odesenfans:
Problem: the broadcast function does not give enough information on the status of the broadcast operation. Some logs are printed but there is no way for the user to determine the status of the message processing + broadcast programmatically.

Solution: refactor the broadcast function so that:
1. it now returns the status of the message, if successful (pending or processed)
2. it raises a `BroadcastError` exception if the broadcast operation failed for some reason.

Renamed `broadcast` to `_broadcast` as it is now considered a private function.

### DIFF
--- a/examples/httpgateway.py
+++ b/examples/httpgateway.py
@@ -38,12 +38,12 @@ async def source_post(request):
         if secret != app['secret']:
             return web.json_response({'status': 'error',
                                       'message': 'unauthorized secret'})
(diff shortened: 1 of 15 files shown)


MHHukiewitz (approved):
This PR combines a few things besides changing the broadcast function, which could've been put in a different PR, but LGTM.

Important to note that this PR contains breaking changes, as all the `create_` functions now return a tuple.

PR #140, opened 2023-01-11
Internal: user session class

odesenfans:
Problem: we often pass 2-3 objects to each function that represent information on how to connect to the Aleph network:

* the user account
* the API server URL
* the aiohttp session.

This can be simplified to improve the user experience and reduce the complexity of the SDK.

Solution: introduce a new state class: `UserSession`. This object is now passed as the first parameter of every public function. It is initialized with the user account and the API server URL. It is in charge of configuring and managing the aiohttp session object.

The typical usage of the library now looks like this:

```
account = load_my_account()
api_server = "https://aleph.cloud"  # example
async with UserSession(account=account, api_server=api_server) as session:
  message, status = await create_post(
    session=session,
    ...
  )
```

This enables the following improvements:

* Less clutter in function signatures: all public SDK functions now have 1 or 2 fewer arguments.
* The API server URL is now managed when initializing the aiohttp session inside the user session object. Implementations can simply specify the endpoint URL. Ex: `f"{api_server}/api/v0/messages.json` can now be expressed as `"/api/v0/messages.json"`.

Breaking changes:

* The signatures of all public functions of the SDK have been modified. The user must now initialize the user session object and pass it as parameter.

### DIFF
--- a/tests/integration/itest_forget.py
+++ b/tests/integration/itest_forget.py
@@ -4,6 +4,7 @@ import pytest
 
 from aleph_client.asynchronous import create_post, get_posts, get_messages, forget
 from aleph_client.types import Account
+from aleph_client.user_session import UserSession
 from .config import REFERENCE_NODE, TARGET_NODE, TEST_CHANNEL
 from .toolkit import try_until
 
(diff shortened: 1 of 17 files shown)


hoh:
While this makes a lot of sense for functions that publish messages, this looks more cumbersome for just getting messages.

Instead of having to pass a mandatory everywhere, why not make the functions methods of a class instead ?

```python
with AlephClient(api_server=api_server) as client:
  message, status = await client.get_messages(
    ...
  )
```

```python
account = load_my_account()
with AlephClient(account=account, api_server=api_server) as client:
  message, status = await client.create_post(
    ...
  )
```

tests/integration/itest_forget.py
```diff
@@ -41,13 +44,13 @@ async def wait_matching_posts(
 
     post_hash = post_message.item_hash
     reason = "This well thought-out content offends me!"
```
MHHukiewitz, line 46: lol

MHHukiewitz (changes requested):
When testing, I get these errors:
```
============================================================ short test summary info
FAILED tests/unit/test_asynchronous_get.py::test_get_posts - TypeError: __init__() got an unexpected keyword argument 'base_url'
FAILED tests/unit/test_asynchronous_get.py::test_get_messages - TypeError: __init__() got an unexpected keyword argument 'base_url'
FAILED tests/unit/test_vm_cache.py::test_local_vm_cache - TypeError: __init__() got an unexpected keyword argument 'base_url'
ERROR tests/unit/test_synchronous_get.py::test_get_posts
=============================================== 3 failed, 29 passed, 9 warnings, 1 error in 1.35s

PR #143, opened 2023-01-11
Fix: remove aiohttp from TestVmCache __init__()

odesenfans:
Problem: the aiohttp session is not required in this object. Solution: remove it from the base class constructor and use it only for the real VM cache.

### DIFF
--- a/src/aleph_client/vm/cache.py
+++ b/src/aleph_client/vm/cache.py
@@ -21,8 +21,6 @@ def sanitize_cache_key(key: str) -> CacheKey:
 class BaseVmCache(abc.ABC):
     """Virtual Machines can use this cache to store temporary data in memory on the host."""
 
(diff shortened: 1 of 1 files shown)


src/aleph_client/vm/cache.py
```diff
@@ -23,9 +23,6 @@ class BaseVmCache(abc.ABC):
 
     session: ClientSession
```
MHHukiewitz, line 24: should also be moved to VmCache

MHHukiewitz (changes requested):
Otherwise, LGTM

src/aleph_client/vm/cache.py
odesenfans: Fixed.

MHHukiewitz (approved):
LGTM

PR #144, opened 2023-01-12
Internal: enable type checking for examples

odesenfans:
Problem: the examples are not checked with mypy and contain incorrect code.

Solution: add them to the list of files checked with mypy, fix the existing issues. While we're at it, remove commented code.


odesenfans:
Duplicate of #142, didn't see it was merged.

PR #159, opened 2023-05-17
Fix: generate valid signatures with Python 3.11

odesenfans:
Problem: signatures generated with Python 3.11 are invalid. The signature buffer is affected by the new formatting of string enums.

Solution: use the value of the message type enum.


hoh:
This fix should go in aleph-sdk-python.

Replaced with https://github.com/aleph-im/aleph-sdk-python/pull/30

PR #164, opened 2023-07-08
Add sign message command

MHHukiewitz:
Allows users to pass JSON strings to the Aleph CLI to be signed as Aleph message (requiring the properties `chain`, `type` and `item_hash`):
```sh
aleph message sign --message "{\"chain\":\"SupportedChains.Ethereum\",\"sender\":\"0x48A87924135892bEE2F264547a0D69909b3bA5C6\",\"type\":\"authorization_challenge\",\"item_hash\":\"d5b64a7c115e477a1e292e937143c1e7bfcf5442f43423f639eb8d9f57a91c76b776e90cc53ba1e7150a5d7521a428b721a101427e37f3d00a6c3cd124cf118b\"}"
```
At a later point this command should also allow signing of arbitrary strings, which is a nice utility.


MHHukiewitz:
Kind of weird, I don't understand why the tests in the Docker containers fail, they all pass for me locally.

hoh:
> Kind of weird, I don't understand why the tests in the Docker containers fail, they all pass for me locally.

Did you install the same version of the testing tools ? Most of our CI downloads the latest version, which is subject to updates.

src/aleph_client/commands/account.py
```diff
@@ -115,3 +117,37 @@ def export_private_key(
...
+    setup_logging(debug)
+
+    if private_key is not None:
+        private_key_file = None
+    elif private_key_file and not os.path.exists(private_key_file):
+        exit(0)
+
+    if message is None:
+        # take from stdin
+        message = ''.join(sys.stdin.readlines())
```
hoh, line 144: Why do you join lines ?

src/aleph_client/commands/account.py
```diff
@@ -115,3 +117,37 @@ def export_private_key(
 def path():
     if sdk_settings.PRIVATE_KEY_FILE:
         typer.echo(sdk_settings.PRIVATE_KEY_FILE)
+
+
+@app.command()
+def sign(
```
hoh, line 123: What about calling this `sign_bytes` or something similar, to express that it is different than signing a message ?

src/aleph_client/commands/message.py
```diff
@@ -262,3 +264,34 @@ def watch(
...
+    """Sign an aleph message with a private key. If no --message is provided, the message will be read from stdin."""
+
+    if private_key is not None:
+        private_key_file = None
+    elif private_key_file and not os.path.exists(private_key_file):
+        exit(0)
+
+    if message is None:
+        # take from stdin
+        message = ''.join(sys.stdin.readlines())
```
hoh, line 288: What's the point of removing `\n` if it's going to be `json.load()`-ed later ?

src/aleph_client/commands/message.py
```diff
@@ -262,3 +264,34 @@ def watch(
...
+
+    account: AccountFromPrivateKey = _load_account(private_key, private_key_file)
+    loop = asyncio.get_event_loop()
+    try:
+        coroutine = account.sign_message(json.loads(message))
+        signed_message = loop.run_until_complete(coroutine)
```
hoh, line 294: Can we use `asyncio.run(...)` instead ?

tests/unit/test_commands.py
```diff
@@ -90,3 +90,35 @@ def test_message_find():
...
+            "message",
+            "sign",
+            "--message",
+            "{\"chain\":\"SupportedChains.Ethereum\",\"sender\":\"0x48A87924135892bEE2F264547a0D69909b3bA5C6\",\"type\":\"authorization_challenge\",\"item_hash\":\"d5b64a7c115e477a1e292e937143c1e7bfcf5442f43423f639eb8d9f57a91c76b776e90cc53ba1e7150a5d7521a428b721a101427e37f3d00a6c3cd124cf118b\"}",
+        ],
+        capture_output=True,
+    )
+
+    assert result.returncode == 0
+    assert b"signature" in result.stdout
```
hoh, line 108: Can you add tests that uses stdin ?

src/aleph_client/commands/account.py
MHHukiewitz: That's the only way to get all the lines from stdin TextIO. Should I instead join them with `\n`?

MHHukiewitz:
@hoh Tests are passing now, suggestions have been committed :+1:

Issue #87, opened 2022-08-17
Test coverage is only 40%


hoh:
## Context

As of today ( f0ea50fc3418fab2955c7f9c699efe2b4e1fb398 ), unit test coverage of the project is only 40%.

Using `pytest --cov` runs tests while displaying the coverage.

## Problem

This gives us low confidence about new commits not breaking something.

## Proposal

Add unit test that focus on increasing the coverage, especially for chain support.

PR #170, opened 2023-10-03
Internal: Add Unit Test

1yam:
Problem:
 - Some feature of Aleph-client doesn't have unit tests and https://github.com/aleph-im/aleph-client/issues/87

Solutions:
 - Add Unit test for aleph file upload
 - Add Unit test for aleph account path
 
 Current Coverage : 45 %

### DIFF
--- a/tests/unit/test_commands.py
+++ b/tests/unit/test_commands.py
@@ -84,6 +84,7 @@ def test_message_find():
 
 
 def test_file_upload():
+    #  Test upload a file to aleph network by creating a file and upload it to an aleph node
     with NamedTemporaryFile() as temp_file:
         temp_file.write(b"Hello World \n")
         result = subprocess.run(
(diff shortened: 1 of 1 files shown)


tests/unit/test_commands.py
```diff
@@ -36,6 +37,13 @@ def test_account_export_private_key(account_file: Path):
     assert len(result.stdout.strip()) == 66
 
 
+def test_account_path():
```
hoh, line 40: What is this testing ?

tests/unit/test_commands.py
```diff
@@ -73,3 +81,19 @@ def test_message_find():
         b"bd79839bf96e595a06da5ac0b6ba51dea6f7e2591bb913deccded04d831d29f4"
         in result.stdout
     )
+
+
+def test_file_upload():
```
hoh, line 86: Can you add a docstring that explains what this is testing ?

tests/unit/test_commands.py
1yam: This test check that path is set well

MHHukiewitz:
Is superseded by #173

PR #173, opened 2023-11-10
Use latest post-v0.7.0 SDK - fix breaking changes

MHHukiewitz:
PR #164, which has recently been merged, used the latest main branch of `aleph-sdk-python`. There have been further breaking changes since the last CI/CD run of that PR, so a broken branch has mistakenly been merged into this repo's master branch.

This PR solves multiple issues:
- Use new `MessageFilter` and `PostFilter` properties and other changes to the `AlephClient` and `AuthenticatedAlephClient` interfaces.
- Introduce an `AsyncTyper` class to accommodate for the fact that `aleph-sdk-python` removed sync wrappers for the client classes.
- Fixes a MacOS CI/CD issue, forcing it to use Python 3.11 instead of 3.12.

### DIFF
--- a/tests/unit/test_commands.py
+++ b/tests/unit/test_commands.py
@@ -1,9 +1,8 @@
 import json
-import subprocess
 from pathlib import Path
-from aleph.sdk.chains.ethereum import ETHAccount
 from tempfile import NamedTemporaryFile
 
+from aleph.sdk.chains.ethereum import ETHAccount
 from typer.testing import CliRunner
 
 from aleph_client.__main__ import app
@@ -55,9 +54,7 @@ def test_account_export_private_key(account_file: Path):
 
 
 def test_account_path():
-    result = runner.invoke(
-        app, ["account", "path"]
-    )
+    result = runner.invoke(app, ["account", "path"])
     assert result.stdout.startswith("/")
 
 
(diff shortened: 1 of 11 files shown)


tests/unit/test_commands.py
```diff
@@ -52,6 +54,13 @@ def test_account_export_private_key(account_file: Path):
...
 
+def test_account_path():
+    result = runner.invoke(
+        app, ["account", "path"]
+    )
+    assert result.stdout.startswith("/")
+
+
 def test_message_get():
     # Use subprocess to avoid border effects between tests caused by the initialisation
```
1yam, line 65: Maybe don't need to use subprocess anymore ? if we use aynsc ?

tests/unit/test_commands.py
```diff
@@ -143,3 +152,20 @@ def test_sign_raw_stdin():
 
     assert result.returncode == 0
     assert b"0x" in result.stdout
+
+
+def test_file_upload():
+    #  Test upload a file to aleph network by creating a file and upload it to an aleph node
+    with NamedTemporaryFile() as temp_file:
+        temp_file.write(b"Hello World \n")
+        result = subprocess.run(
```
1yam, line 161: same here no ?

1yam (approved):
LGTM

PR #175, opened 2023-11-13
Feature: download File

1yam:
Solutions:

Create a new command that allows users to download a file from a hash.

### DIFF
--- a/src/aleph_client/commands/files.py
+++ b/src/aleph_client/commands/files.py
@@ -3,7 +3,7 @@ from pathlib import Path
 from typing import Optional
 
 import typer
-from aleph.sdk import AuthenticatedAlephHttpClient
+from aleph.sdk import AlephHttpClient, AuthenticatedAlephHttpClient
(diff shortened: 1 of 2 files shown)


src/aleph_client/commands/files.py
```diff
@@ -99,3 +99,37 @@ async def upload(
...
+    output_file_path = output_path / f"{file_name}{file_extension}"
+
+    async with AlephHttpClient(api_server=sdk_settings.API_HOST) as client:
+        logger.info(f"Downloading {hash} ...")
+        async with AlephHttpClient(api_server=sdk_settings.API_HOST) as client:
+            with open(output_file_path, "wb") as fd:
+                if use_ipfs:
+                    await client.download_file_to_buffer(hash, fd)
+                else:
+                    await client.download_file_to_buffer(hash, fd)
```
MHHukiewitz, line 133: This looks like it's doing exactly the same, no matter how

PR #178, opened 2023-11-25
Create Instance

MHHukiewitz:
Adds two new functions:
-`aleph instance create`: Creates a new instance
-`aleph instance delete`: Deletes/Forgets an instance


1yam (approved):
LGTM

PR #181, opened 2023-11-29
Replace all usages of `requests` module with `aiohttp`

MHHukiewitz:
As we are now using `AsyncTyper`, we can freely use async `aiohttp`-style requests everywhere


1yam (approved):
LGTM

PR #180, opened 2023-11-27
Upgrade aleph-message to version 0.4.1

MHHukiewitz:

PR #182, opened 2023-11-29
Aliel fix nuls2 deps

aliel:
puls2-python is broken
change dependency to aleph-nuls2


MHHukiewitz:
Needs to be rebased to fix the dependency conflict on aleph-message (see #180)

PR #185, opened 2023-12-05
Fix: Importing account from private key wasn't possible

1yam:
### **_Problem:_**
1. **Wasn't allowed to simply send a private key**
2. **Encoding on file wasn't correct**


MHHukiewitz (approved):
LGTM

PR #188, opened 2023-12-12
Fix: Amend Function

1yam:
Problem: 
- The message wasn't sent as an amend.
- The time sent corresponds to the original message time and not current.


MHHukiewitz (approved):
LGTMeeee

PR #192, opened 2023-12-15
Improve Instance Creation and Add `instances list`

MHHukiewitz:
Improvements:
- Use `rich` more broadly in instances and `utils.py`
- Add `aleph instance list` command
- Improve `aleph instance create` to be easier to understand and more interactive
- Add fresh Ubuntu 22, Debian 11 and Debian 12 rootfs item hashes to choice


1yam (approved):
LGTM

PR #196, opened 2024-01-30
Add deprecation messages and update docs

MHHukiewitz:
- Raising ImportError for common imports like `aleph_client.AlephClient` or `aleph_client.asynchronous`
- Update docs for aggregate and posts

### DIFF
--- a/docs/content/aggregates.rst
+++ b/docs/content/aggregates.rst
@@ -6,21 +6,34 @@ Aggregates
 
 Aggregates are a key-value store specific to an account.
 Each time a new aggregate message is received for a specific account, the
-nodes add a new layer on top and mutate the global storage for this account.
+nodes update the aggregate for this account.
+
+Like a dictionary update, if a key already exists, it is updated,
+otherwise it is created.
 
 Query aggregate of an account
 -----------------------------
 
-To query keys from an account aggregate, you need to call the fetch
-aggregate function.
+To query keys from an account aggregate, you need to call the
+fetch_aggregate function on the client.
 
-Synchronous version:
+Since version 0.8.0, only the asynchronous methods are available.
+To use them in the Python REPL, simply start it with:
+
+.. code-block:: bash
+
+    $ python3 -m asyncio
+
+Then you can use the asynchronous methods:
 
 .. code-block:: python3
 
-    >>> from aleph_client.synchronous import fetch_aggregate
-    >>> fetch_aggregate("0x06DE0C46884EbFF46558Cd1a9e7DA6B1c3E9D0a8",
-    ... "profile")
+    >>> from aleph.sdk.client import AlephHttpClient
+    >>> async with AlephHttpClient() as client:
+    ...     await client.fetch_aggregate(
+    ...         "0x06DE0C46884EbFF46558Cd1a9e7DA6B1c3E9D0a8",
+    ...         "profile",
+    ...     )
     {"bio": "tester", "name": "Moshe on Ethereum"}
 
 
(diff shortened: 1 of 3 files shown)


docs/content/aggregates.rst
```diff
@@ -6,21 +6,34 @@ Aggregates
...
+To use them in the Python REPL, simply start it with:
+
+.. code-block:: bash
+
+    $ python3 -m asyncio
 
-Synchronous version:
+Then you can use the asynchronous methods:
 
 .. code-block:: python3
```
hoh, line 29: Do you need to use `python -m asyncio` to have access to `await` in the REPL ? If so, should that be documented ? Or the altenative using `asyncio.run(coroutine)` ?

src/aleph_client/__init__.py
```diff
@@ -8,3 +8,12 @@
...
     del get_distribution, DistributionNotFound
+
+# Deprecation check
+moved_types = ["AlephClient", "AuthenticatedAlephClient", "synchronous", "asynchronous"]
+
+
+def __getattr__(name):
+    if name in moved_types:
+        raise ImportError(f"The 'aleph_client.{name}' type is deprecated and has been removed from aleph_client. Please use `aleph.sdk.{name}` instead.")
+    raise AttributeError(f"module 'aleph_client' has no attribute '{name}'")
```
hoh, line 19: Should we instead call the parent implementation ? 
```suggestion
    return super().__getattr__(name)
```

docs/content/aggregates.rst
MHHukiewitz: Yeah, that's the case, I documented it just above

MHHukiewitz: Because the docs were centered around REPL use, I thought I just keep it that way. If people want to know more about async, they can look it up themselves.

src/aleph_client/__init__.py
MHHukiewitz: Absolutely.

hoh:
`ImportError` does not seem like the right exception here :thinking: 
[See https://docs.python.org/3/library/exceptions.html](https://docs.python.org/3/library/exceptions.html#exception-hierarchy)
`AttributeError`  is the standard class for this in Python, see 
https://docs.python.org/3/reference/datamodel.html#object.__getattr__

nesitor (approved):

nesitor:
LGTM, but there are tests that are failing, and we need to fix those before merge it.

src/aleph_client/__init__.py
```diff
@@ -8,3 +8,12 @@
...
 finally:
     del get_distribution, DistributionNotFound
+
+# Deprecation check
+moved_types = ["AlephClient", "AuthenticatedAlephClient", "synchronous", "asynchronous"]
+
+
+def __getattr__(name):
+    if name in moved_types:
+        raise ImportError(f"The 'aleph_client.{name}' type is deprecated and has been removed from aleph_client. Please use `aleph.sdk.{name}` instead.")
```
hoh, line 18: `ImportError` does not seem like the right exception here :thinking: 
[See https://docs.python.org/3/library/exceptions.html](https://docs.python.org/3/library/exceptions.html#exception-hierarchy)
`AttributeError`  is the standard class for this in Python, see 
https://docs.python.org/3/reference/datamodel.html#object.__getattr__

MHHukiewitz: But actually, ImportError is fine: https://docs.python.org/3/library/exceptions.html#ImportError

hoh: I don't agree, this is not an import error. It is not "the [import](https://docs.python.org/3/reference/simple_stmts.html#import) statement has troubles trying to load a module"

PR #199, opened 2024-03-13
Implement hypervisor field for instances

nesitor:
Problem: We cannot select the hypervisor to publish an instance, and the memory by default isn't the standard.

Solution: Add hypervisor field using firecracker by default and specify the default memory for instances.

### DIFF
--- a/src/aleph_client/commands/instance.py
+++ b/src/aleph_client/commands/instance.py
@@ -183,7 +183,6 @@ async def create(
                 sync=True,
                 rootfs=rootfs,
                 rootfs_size=rootfs_size,
-                rootfs_name="rootfs",
                 storage_engine=StorageEnum.storage,
                 channel=channel,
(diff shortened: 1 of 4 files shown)


src/aleph_client/commands/instance.py
```diff
@@ -179,14 +183,15 @@ def validate_ssh_pubkey_file(file: Union[str, Path]) -> Path:
                 sync=True,
                 rootfs=rootfs,
                 rootfs_size=rootfs_size,
-                rootfs_name=rootfs_name,
+                rootfs_name="rootfs",
```
MHHukiewitz, line 186: I thought in the SDK PR this parameter has been removed?

MHHukiewitz (commented):
LGTM, just one question before merging

MHHukiewitz (approved):
LGTM

PR #202, opened 2024-04-17
Add SPA field on domains

nesitor:
Problem: SPA application needs special settings on DNS resolution service to be loaded well.

Solution: Add the SPA field on the domain aggregate to note that is an SPA application.


MHHukiewitz (approved):
It is a mess to review as the similar functions swapped places, but LGTM

aliel (approved):

gmolki (approved):

PR #208, opened 2024-05-22
207 adding creation of an instance with payg

Antonyjin:
- [x] Use of a constant instead of None to indicate the end of processing
- [x] Use a structrure type instead of a dictionary
- [x] Specify caught exceptions
- [x] Put the MachineUsage -> array list transformation into a function
- [x] Pass Structure to the queue and specify its type as an annotation
- [x] Run mypy and ruff and fix problems encountered
- [ ] Move instance-related code to another file (refactor page in general)
- [ ] Add unit tests + Learn more about TDD: Test-Driven Development
- [x] Add docstrings
- [x] Check presence of type annotations (mypy can help)
- [x] Change the hardcoded string about version


src/aleph_client/commands/utils.py
```diff
@@ -200,3 +206,111 @@ def is_environment_interactive() -> bool:
...
+    try:
+        async with session.get(node['address'], timeout=timeout) as resp:
+            if resp.status == 200:
+                    html_content = await resp.text()
+                    soup = BeautifulSoup(html_content, 'html.parser')
+                    p_tags = soup.find_all('p')
+                    for p in p_tags:
+                        if 'Running version' in p.text:
+                            version = p.find('i').text
+                            return version
```
hoh, line 312: You may want to use the last `server` header in the response instead.

See this code from the metrics: https://github.com/aleph-im/aleph-scoring/blob/2444ce57c6ad5a6693be29675b451ddff0b3dfc4/aleph_scoring/metrics/__init__.py#L172-L190

src/aleph_client/commands/utils.py
```diff
@@ -200,3 +206,111 @@ def is_environment_interactive() -> bool:
...
+    await queue.put(None)
+
+async def fetch_and_queue(session: aiohttp.ClientSession, node: NodeInfo, queue: asyncio.Queue, progress: Progress, task: list, item_hashes: list):
+    try:
+        system_info, version = await asyncio.gather(
+            fetch_crn_system(session, node),
+            get_crn_version(session, node)
+        )
+        async with session.get(node["address"] + "status/check/ipv6", timeout=10) as resp:
+            if resp.status == 200:
```
hoh, line 265: What about using `resp.raise_for_status()` ?

src/aleph_client/commands/utils.py
```diff
@@ -200,3 +206,111 @@ def is_environment_interactive() -> bool:
...
+        data = await queue.get()
+        if data is None:
+            break
+        table.add_row(*data)
+
+async def fetch_crn_system(session: aiohttp.ClientSession, node: NodeInfo):
+    try:
+        async with session.get(node["address"] + "about/usage/system", timeout=10) as resp:
+            if resp.status == 200:
+                data = await resp.json()
```
hoh, line 294: Would it make sense to use the model `CpuProperties` from aleph-message ?

```suggestion
                data_raw = await resp.json()
                data = CpuProperties.parse_obj(data_raw)
```

See https://github.com/aleph-im/aleph-message/blob/3c0144adb73d54d146cb6ce407a6bf5a14cf7995/aleph_message/models/execution/environment.py#L71

src/aleph_client/commands/instance.py
```diff
@@ -178,6 +199,9 @@ def validate_ssh_pubkey_file(file: Union[str, Path]) -> Path:
     async with AuthenticatedAlephHttpClient(
         account=account, api_server=sdk_settings.API_HOST
     ) as client:
+        payment = None
```
hoh, line 202: Can you add a typing annotation ?

src/aleph_client/commands/utils.py
```diff
@@ -200,3 +206,111 @@ def is_environment_interactive() -> bool:
...
+            fetch_task = asyncio.create_task(fetch_data(session, node_info, queue, progress, task, item_hashes))
+            update_task = asyncio.create_task(update_table(queue, table))
+            await asyncio.gather(fetch_task, update_task)
+
+    return item_hashes
+
+async def fetch_data(session: aiohttp.ClientSession, node_info: NodeInfo, queue: asyncio.Queue, progress: ProgressTable, task: list, item_hashes: list):
+    tasks = [fetch_and_queue(session, node, queue, progress, task, item_hashes) for node in node_info.nodes]
+    await asyncio.gather(*tasks)
+    await queue.put(None)
```
hoh, line 256: What about https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue.task_done ?

src/aleph_client/commands/utils.py
```diff
@@ -200,3 +206,111 @@ def is_environment_interactive() -> bool:
...
+        
+        pass
+    finally:
+        progress.update(task, advance=1)
+
+async def update_table(queue: asyncio.Queue, table: Table):
+    while True:
+        data = await queue.get()
+        if data is None:
+            break
```
hoh, line 287:
```suggestion
    while not queue.task_done():
        data = await queue.get()
```

src/aleph_client/commands/utils.py
```diff
@@ -200,3 +206,111 @@ def is_environment_interactive() -> bool:
...
+    progress_table = ProgressTable(progress, table)
+    item_hashes = []
+
+    queue = asyncio.Queue()
+
+    async with aiohttp.ClientSession() as session:
+        with Live(progress_table, console=console, refresh_per_second=2):
+            fetch_task = asyncio.create_task(fetch_data(session, node_info, queue, progress, task, item_hashes))
+            update_task = asyncio.create_task(update_table(queue, table))
+            await asyncio.gather(fetch_task, update_task)
```
hoh, line 249: Avoid using `asyncio.create_task()`, if an exception happens some tasks will still be running.
Instead, use `asyncio.gather` on the coroutines directly:

```suggestion
            await asyncio.gather((
                fetch_data(session, node_info, queue, progress, task, item_hashes),
                update_table(queue, table),
            )
```

hoh (commented):
Nice start. 

I added comments to improve the structure of the code.

Can you add type annotations and check the output of `mypy` and `ruff check` on the new code ?

src/aleph_client/commands/utils.py
Antonyjin: That is a best way to do it indeed, i modified it.

src/aleph_client/commands/utils.py
Antonyjin: Done!

src/aleph_client/commands/utils.py
Antonyjin: In my mind, it is not necessary to use this function because `asyncio.gather` already handles task synchronization and completion effectively.
Adding these functions would maybe increase complexity without providing a lot of benefit.

src/aleph_client/commands/utils.py
```diff
@@ -223,6 +228,64 @@ def __rich_console__(self, console, options):
...
+    
+class CoreFrequencies(BaseModel):
+    min: float
+    max: float
+
+    @classmethod
+    def from_psutil(cls, psutil_freq: psutil._common.scpufreq):
+        min_ = psutil_freq.min or psutil_freq.current
+        max_ = psutil_freq.max or psutil_freq.current
+        return cls(min=min_, max=max_)
```
hoh, line 254: Why does `aleph-client` need to know the CPU load average and frequencies of the user's machine ?

src/aleph_client/commands/utils.py
```diff
@@ -274,13 +340,20 @@ async def fetch_and_queue(session: aiohttp.ClientSession, node: NodeInfo, queue:
...
+                    ram = "TESTING"
                     await queue.put((score, node_name, cpu, ram, hdd, version, node_stream, node_address))
                     item_hashes.append(node_stream)
+    except TimeoutError:
+        logger.debug(f'Timeout while fetching: {url}')
+    except aiohttp.client_exceptions.ClientConnectionError:
+        logger.debug(f'Error on connection: {url}')
     except Exception as e:
-        pass
+        logger.debug(f'This error occured: {e}')
```
hoh, line 356:
```suggestion
        logger.error(f'This error occured: {e}')
```

src/aleph_client/commands/utils.py
```diff
@@ -291,28 +364,42 @@ async def update_table(queue: asyncio.Queue, table: Table):
             break
         table.add_row(*data)
 
-async def fetch_crn_system(session: aiohttp.ClientSession, node: NodeInfo) -> str:
+async def fetch_crn_system(session: aiohttp.ClientSession, node: NodeInfo) -> CpuProperties:
+    data = None
```
hoh, line 368:
```suggestion
    data: Optional[MachineUsage] = None
```

src/aleph_client/commands/utils.py
```diff
@@ -291,28 +364,42 @@ async def update_table(queue: asyncio.Queue, table: Table):
...
+                data_raw = await resp.json()
+                data = MachineUsage.parse_obj(data_raw)
+                # print("data:", data)
+    except TimeoutError:
+        logger.debug(f'Timeout while fetching: {url}')
+    except aiohttp.client_exceptions.ClientConnectionError:
+        logger.debug(f'Error on connection: {url}')
     except Exception as e:
-        data = None
+        logger.debug(f'This error occured: {e}')
```
hoh, line 383: It is usually a bad practice to catch every exception, unless the context is a service that must keep working. 

In this case, two known exceptions are already handled and other issues (memory full, ...) may not be handled correctly by the program anyways.
What if it is a "variable not defined" for example ? This should definitely not go unnoticed.

src/aleph_client/commands/utils.py
```diff
@@ -291,28 +364,42 @@ async def update_table(queue: asyncio.Queue, table: Table):
...
     return data
 
 async def get_crn_version(session: aiohttp.ClientSession, node: NodeInfo) -> str:
-    version = None
+    url = node["address"]
+    version = "Can't fetch the version"
```
hoh, line 388: Why use a hardcoded string instead of either returning `Optional[str]` or raise an exception ?

Antonyjin: This was a temporary state for the version, I just wanted to check if the fetch of the version worked, and the first thing that came to mind was to put this.

My code needs something in the version field to be able to put it in the tab. So raising an exception, or putting it as optional, wouldn't work for the current version.

I intend to change it in the future,

hoh:
Replaced with https://github.com/aleph-im/aleph-client/pull/216

PR #218, opened 2024-06-26
feat(network): more verbose message error for better debugging

Psycojoker:


hoh (approved):

Psycojoker:
Let's merge this since it's only the #223 problem here that is unrelated.

PR #234, opened 2024-07-24
Problem: Hypervisor passed in parameter was not respected

olethanh:
if an user passed the --hypervisor option, the question was still asked


1yam (approved):
LGTM, just codecov not happy of 0.02% diff

src/aleph_client/commands/instance/__init__.py
```diff
@@ -24,6 +24,7 @@
 from aleph_message.models.execution.base import Payment, PaymentType
 from aleph_message.models.execution.environment import HypervisorType
 from aleph_message.models.item_hash import ItemHash
+from click import echo
```
Psycojoker, line 27: This looks like an unused import?

Psycojoker (approved):
Apart from the unused import LGTM

PR #236, opened 2024-07-31
Fix: Ensure consistent URL formatting in _get_ipv6_address

1yam:
Problem:

In the node_list, the addresses are stored inconsistently. Some nodes have addresses without a trailing slash, like:
``` json
"address": "https://acn-000060.nodeforge.io"
```
While others have addresses with a trailing slash:

``` json
"address": "https://aleph3.serverrg.eu/"
```
Solution:

Ensure consistent URL formatting by appending the path correctly, regardless of whether the address ends with a slash. This will prevent any issues caused by missing or double slashes when constructing URLs. The implementation will check if the last character of the address is a slash and adjust accordingly.


philogicae (approved):
LGTM

PR #237, opened 2024-07-31
Feature: prompt for payment type

1yam:
Problem:

We currently blocking user from creating Confidential vm using hold method. However, in the case user use NFT he need to set the payment type to Hold.

### DIFF
--- a/src/aleph_client/commands/instance/__init__.py
+++ b/src/aleph_client/commands/instance/__init__.py
@@ -78,7 +78,7 @@ async def create(
         Path("~/.ssh/id_rsa.pub").expanduser(),
         help="Path to a public ssh key to be added to the instance.",
     ),
-    print_messages: bool = typer.Option(True),
+    print_messages: bool = typer.Option(False),
     rootfs: str = typer.Option(
         "Ubuntu 22",
         help="Hash of the rootfs to use for your instance. Defaults to Ubuntu 22. You can also create your own rootfs and pin it",
--- a/src/aleph_client/commands/instance/__init__.py
+++ b/src/aleph_client/commands/instance/__init__.py
@@ -58,7 +58,7 @@ app = AsyncTyper(no_args_is_help=True)
 @app.command()
 async def create(
     hold: bool = typer.Option(
-        default=False,
+        default=None,
         help="Pay using the holder tier instead of pay-as-you-go",
     ),
     channel: Optional[str] = typer.Option(default=None, help=help_strings.CHANNEL),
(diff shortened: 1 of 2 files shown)


hoh:
Can you ensure that tests cover the changes ?

src/aleph_client/commands/instance/__init__.py
```diff
@@ -78,7 +78,7 @@ async def create(
         Path("~/.ssh/id_rsa.pub").expanduser(),
         help="Path to a public ssh key to be added to the instance.",
     ),
-    print_messages: bool = typer.Option(False),
+    print_messages: bool = typer.Option(True),
```
Psycojoker, line 81: That looks like an unrelated change to this PR

src/aleph_client/commands/instance/__init__.py
```diff
@@ -284,19 +295,17 @@ def validate_ssh_pubkey_file(file: Union[str, Path]) -> Path:
         item_hash: ItemHash = message.item_hash
 
         console = Console()
-        if crn and (confidential or not hold):
+        if crn and not hold:
             if not crn.url:
                 return
-            async with AuthenticatedAlephHttpClient(account=account, api_server=sdk_settings.API_HOST) as client:
```
Psycojoker, line 290: Why did you removed this first switch statement? It doesn't useful but I'm curious.

src/aleph_client/commands/instance/__init__.py
```diff
@@ -284,19 +295,17 @@ def validate_ssh_pubkey_file(file: Union[str, Path]) -> Path:
...
-                        print(status, result)
-                        echo(f"Could not start instance  {item_hash} on CRN")
-                        return
+            account = _load_account(private_key, private_key_file)
+            async with VmClient(account, crn.url) as crn_client:
+                status, result = await crn_client.start_instance(vm_id=item_hash)
+                logger.debug(status, result)
+                if int(status) != 200:
+                    print(status, result)
+                    echo(f"Could not start instance  {item_hash} on CRN")
```
Psycojoker, line 307:
```suggestion
                    echo(f"Could not start instance {item_hash} on CRN")
```

src/aleph_client/commands/instance/__init__.py
1yam: I did remove it cause it was here but no used

1yam:
> Can you ensure that tests cover the changes ?

We currently don't have any tests for instance creation in the aleph-client. It would be interesting to create some, but how should we approach this? Aleph-client is just a CLI for the SDK, so how do we test it?

And maybe need another PR for that

philogicae (approved):

Issue #245, opened 2024-08-14
Could not create an instance due to Invalid cross-device link


hoh:
```
(venv) user@sandbox ~/R/a/aleph-client (various-fixes-after-tests) [0|127]> aleph instance confidential-init-session 79724e4cba21ac034b1877e83b612db05edaf82aa8a66ee044fd788bfca8d00d https://dchq.staging.aleph.sh
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/user/Repos/aleph-im/aleph-client/src/aleph_client/utils.py:74 in runner                   │
│                                                                                                  │
│   71 │   │   │                                                                                   │
│   72 │   │   │   @wraps(f)                                                                       │
│   73 │   │   │   def runner(*args, **kwargs):                                                    │
│ ❱ 74 │   │   │   │   return asyncio.run(f(*args, **kwargs))                                      │
│   75 │   │   │                                                                                   │
│   76 │   │   │   decorator(runner)                                                               │
│   77 │   │   else:                                                                               │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │   args = ()                                                                                  │ │
│ │      f = <function confidential_init_session at 0x7f45f4fe3a60>                              │ │
│ │ kwargs = {                                                                                   │ │
│ │          │   'vm_id': '79724e4cba21ac034b1877e83b612db05edaf82aa8a66ee044fd788bfca8d00d',    │ │
│ │          │   'domain': 'https://dchq.staging.aleph.sh',                                      │ │
│ │          │   'policy': 1,                                                                    │ │
│ │          │   'private_key': None,                                                            │ │
│ │          │   'private_key_file':                                                             │ │
│ │          PosixPath('/home/user/.aleph-im/private-keys/ethereum.key'),                       │ │
│ │          │   'debug': False                                                                  │ │
│ │          }                                                                                   │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                  │
│ /nix/store/g9sjwndxjirb1k2570hfn902qws6h744-python3-3.12.4/lib/python3.12/asyncio/runners.py:194 │
│ in run                                                                                           │
│                                                                                                  │
│   191 │   │   │   "asyncio.run() cannot be called from a running event loop")                    │
│   192 │                                                                                          │
│   193 │   with Runner(debug=debug, loop_factory=loop_factory) as runner:                         │
[...]

PR #246, opened 2024-08-14
Fix: Cross-device links crashed 'confidential-init-session'

hoh:
Solves https://github.com/aleph-im/aleph-client/issues/245

The function attempts to move a file from `/tmp`
to ~/.aleph-im/confidential_sessions .

This crashed if those were on a different
filesystem.

Replacing `pathlib.Path.rename` with `shutil.move` solves this issue as the later can handle this situation.

### DIFF
--- a/src/aleph_client/commands/instance/__init__.py
+++ b/src/aleph_client/commands/instance/__init__.py
@@ -602,7 +602,8 @@ async def confidential_init_session(
         echo("Could not get the certificate from the CRN.")
         return
 
-    platform_cert_path = Path(platform_file).rename(session_dir / "platform_certificate.pem")
+    # pathlib.Path.rename raises "Invalid cross-device link" if the destionation is not on the current filesystem.
+    platform_cert_path = shutil.move(platform_file, session_dir / "platform_certificate.pem")
     certificate_prefix = str(session_dir) + "/vm"
 
     # Create local session files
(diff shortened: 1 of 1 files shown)


src/aleph_client/commands/instance/__init__.py
```diff
@@ -602,7 +602,8 @@ async def confidential_init_session(
         echo("Could not get the certificate from the CRN.")
         return
 
-    platform_cert_path = Path(platform_file).rename(session_dir / "platform_certificate.pem")
+    # pathlib.Path.rename raises "Invalid cross-device link" if the destionation is not on the current filesystem.
```
olethanh, line 605:
```suggestion
    # pathlib.Path.rename raises "Invalid cross-device link" if the destination is not on the current filesystem.
```

src/aleph_client/commands/instance/__init__.py
```diff
@@ -602,7 +602,8 @@ async def confidential_init_session(
         echo("Could not get the certificate from the CRN.")
         return
 
-    platform_cert_path = Path(platform_file).rename(session_dir / "platform_certificate.pem")
+    # pathlib.Path.rename raises "Invalid cross-device link" if the destination is not on the current filesystem.
+    platform_cert_path = shutil.move(platform_file, session_dir / "platform_certificate.pem")
```
olethanh, line 606: did you check the return type of shutil.move ?

hoh: No, but the code works after the change, and the official doc mentions:
>  Changed in version 3.3: Added explicit symlink handling for foreign filesystems, thus adapting it to the behavior of GNU’s mv. **Now returns dst**.

philogicae (approved):

PR #254, opened 2024-08-21
Fix docs

philogicae:
Wrong imports for account
aleph_client -> aleph.sdk


1yam (approved):
LGTM

olethanh:
are these doc still relevant? I think they have all been moved to aleph-docs repo and the one present here are stall.
If that's the case they should be removed to not be confusing
@philogicae  @hoh

PR #256, opened 2024-08-21
Problem: confidential keep sessions fix

olethanh:
* Keep session prompot was overriding file
 * Initialize should still be called


src/aleph_client/commands/instance/__init__.py
```diff
@@ -824,38 +824,40 @@ async def confidential_init_session(
...
+    # Generate sessions certificate files
+    if not ((session_dir / "vm_godh.b64").exists() and keep_session):
 
-    code, platform_file = await client.get_certificates()
-    if code != 200:
-        echo("Could not get the certificate from the CRN.")
-        return 1
+        code, platform_file = await client.get_certificates()
+        if code != 200:
+            echo("Could not get the certificate from the CRN.")
```
Psycojoker, line 842: Could we have a error message with more information? This really just says "error, handle this shit yourself" and that's not a great UX imo :/ (but idk if we have more information that just the error code)

philogicae (approved):

PR #259, opened 2024-08-22
Fix CRNtable, sort_by keys, autofiltering on reward_add/qemu/coco

philogicae:
image

## Changes

- Cleaning / rewritting / simplification of commands -> instance -> display.py / network.py
- Properly retrieve CRN data
- Fix broken CRNTable
- Add sort_by keys (score, name, version, address, coco, qemu, url)
- Add auto-filtering for reward address, qemu support and confidential support

## How to test

Try to create a PAYG or coco instance without specifying crn_hash and crn_url. It will display the new CRNTable.

- [x] New code does not duplicate the functions of builtin or popular libraries.
- [x] An LLM was used to review the new code and look for simplifications.
- [x] New classes and functions contain docstrings explaining what they provide.
- [ ] All new code is covered by relevant tests.
- [ ] Documentation has been updated regarding these changes.

### DIFF
--- a/src/aleph_client/models.py
+++ b/src/aleph_client/models.py
@@ -117,29 +117,20 @@ class CRNInfo(BaseModel):
 
     @property
     def display_cpu(self) -> str:
-        try:
-            if self.machine_usage:
-                return f"{self.machine_usage.cpu.count:>3}"
-        except:
-            pass
+        if self.machine_usage:
+            return f"{self.machine_usage.cpu.count:>3}"
         return ""
 
     @property
     def display_ram(self) -> str:
-        try:
-            if self.machine_usage:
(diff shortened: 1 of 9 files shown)


olethanh (approved):
Nice it make things a lot smoother.

Only thing I'm not sure is that the 🔒 is clear enough to indicate that this is for confidentials. Maybe in the action menu we could put  "Sort By 🔒 Confidential" ?

src/aleph_client/models.py
```diff
@@ -102,3 +102,56 @@ def from_unsanitized_input(
...
+    version: Optional[str]
+    score: float
+    stream_reward_address: str
+    machine_usage: Optional[MachineUsage]
+    qemu_support: Optional[bool]
+    confidential_computing: Optional[bool]
+
+    @property
+    def display_cpu(self) -> str:
+        try:
```
olethanh, line 120: Don't abuse the `try`, it hide errors (and Rich as is already hide plenty errors)

PR #264, opened 2024-08-29
fix(pyproject): max aleph-sdk-python versiont to respect semver

Psycojoker:
We don't have an upper hand version for our own dependency which mean we can't
safely release a breaking version of aleph-sdk-python because that would break
this package.

## Changes

Set the max version of this dependency to inferior of the next major version.

## How to test

Just install this version with pip (but the CI will do that for you)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ classifiers = [
 ]
 
 dependencies = [
-  "aleph-sdk-python>=1.0.0",
+  "aleph-sdk-python>=1.0.0,<2",
   "setuptools>=65.5.0",
   "aleph-message>=0.4.9",
   "aiohttp==3.9.5",


hoh:
Mypy issue in the CI:
```
src/aleph_client/commands/instance/__init__.py:922:37: error: Argument 1 to "create_session" of "VmConfidentialClient" has incompatible type "str"; expected "ItemHash"  [arg-type]
```

hoh (approved):

hoh:
> Mypy issue in the CI:
> 
> ```
> src/aleph_client/commands/instance/__init__.py:922:37: error: Argument 1 to "create_session" of "VmConfidentialClient" has incompatible type "str"; expected "ItemHash"  [arg-type]
> ```

This is apparently due to changes in the SDK, and would be fixed by publishing a new version of the SDK.

PR #275, opened 2024-09-25
Fixes a lot of the errors raised by ruff

Psycojoker:
Ruff isn't running on this project for various reason and it raises a LOT of errors.

This PR fixes a good part of the errors but not all of them but all this improvements can already be merged so we can benefit from it.

Related ClickUp, GitHub or Jira tickets : ALEPH-127

- [x] All new code is covered by relevant tests.

### DIFF
--- a/src/aleph_client/commands/instance/__init__.py
+++ b/src/aleph_client/commands/instance/__init__.py
@@ -1,13 +1,14 @@
 from __future__ import annotations
 
 import asyncio
+import builtins
 import json
 import logging
 import shutil
 from decimal import Decimal
 from math import ceil
 from pathlib import Path
-from typing import List, Optional, Tuple, Union, cast
+from typing import cast
 
 import aiohttp
 import typer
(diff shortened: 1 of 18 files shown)


src/aleph_client/commands/instance/__init__.py
```diff
@@ -695,17 +709,17 @@ async def list(
             echo(messages.json(indent=4))
         else:
             # Since we filtered on message type, we can safely cast as InstanceMessage.
-            messages = cast(List[InstanceMessage], messages)
+            messages = cast(builtins.list[InstanceMessage], messages)
```
hoh, line 712: I am wondering if we should not update `filter_only_valid_messages` to use [generics](https://docs.python.org/3/library/typing.html#generics) :thinking:

hoh (approved):

src/aleph_client/commands/instance/__init__.py
Psycojoker: I'm not sure it's in the scope of this PR? We probably want to use `Sequence` instead of `list` for it as a type anyway

Psycojoker:
@hoh what version of python do we target? Right now it's stucked on 3.9 and some new notation aren't supported, I want to know if I need to revert them.

Psycojoker:
Outdated

PR #276, opened 2024-09-25
Enable ruff

Psycojoker:
Appart from the weird compilation bug on ubuntu 20.04, the CI failures seems to be already present on master.

ALEPH-175


Psycojoker:
Appart from the weird compilation bug on ubuntu 20.04, the CI failures seems to be already present on master.

philogicae (approved):

PR #242, opened 2024-08-13
Fix: revert upgrade: aiohttp 3.10.2 -> 3.9.5

philogicae:
This new merged PR broke aleph-client on Windows: https://github.com/aleph-im/aleph-client/pull/239
The only fix right now seems to be downgrading to 3.9.5, according to https://github.com/nathom/streamrip/issues/729
image

PR #282, opened 2024-10-01
chore(pyproject): aiohttp respects semver so let's accept all it's patch releases

Psycojoker:
To avoid having to small PRs https://github.com/aleph-im/aleph-client/pull/279

ALEPH-205


Psycojoker:
We have a bug on window for this PR https://github.com/aleph-im/aleph-client/pull/242 which sucks for security reasons

PR #310, opened 2024-12-02
Reduce instance PAYG balance to half

nesitor:
Related ClickUp, GitHub or Jira tickets : [ALEPH-291](https://aleph-im.atlassian.net/browse/ALEPH-291)

- [ ] An LLM was used to review the new code and look for simplifications.
- [X] New classes and functions contain docstrings explaining what they provide.
- [X] All new code is covered by relevant tests.

## Changes

After changing the instance holding price, we need to put the same on the CLI also.
Also, I increased the limit of the disk size and RAM size, accordingly to the new limits.

## How to test

Try to create a normal instance, holding and PAYG and check that the balance required is the half of before.


1yam (approved):
LGTM

PR #331, opened 2025-01-30
Fix unwanted uninit coco vm msg on instance list

philogicae:
Jira: None

- [x] New code does not duplicate the functions of builtin or popular libraries.
- [x] An LLM was used to review the new code and look for simplifications.
- [x] New classes and functions contain docstrings explaining what they provide.
- [x] All new code is covered by relevant tests.

## Changes

- Fix: `aleph instance list` was displaying "Boot uninitialized/unstarted confidential instances with..." not only on coco, but also for all unallocated instances


1yam (approved):
LGTM

PR #336, opened 2025-02-05
migrate test-docker to hatch

Psycojoker:
The test-docker testing job was still using raw commands instead of using hatch.

### DIFF
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -41,8 +41,8 @@ jobs:
 
       - name: Pytest in the Docker image
         run: |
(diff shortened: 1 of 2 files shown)


olethanh:
Don't know much about the test in the client but I would assume it is suppose to not use hatch so it test with the dependencies already installed in the docker and not have hatch reinstall new ones ? 
Also double check if it doesn't increase CI time

Psycojoker:
> Don't know much about the test in the client but I would assume it is suppose to not use hatch so it test with the dependencies already installed in the docker and not have hatch reinstall new ones ?

Not really, it's more that it's the old way of testing before we moved to hatch so I've just adapted it, we just didn't used anything before.

Another solution is to simply remove it but I think still testing the docker build is a good thing.

Psycojoker:
Merged in another way

PR #350, opened 2025-03-04
Fix: Could not import private key in base32/base64

hoh:
The command `aleph account create` did not permit users to import their private key encoded in base32 or base64.

This adds the `--key-format` argument to the command to specify this.


Psycojoker (approved):
LGTM, that's a nice addition!

PR #353, opened 2025-03-14
display scheduler decisions on instance listing

Psycojoker:


Psycojoker:
Added the information from the CRN, still as a POC (they aren't very reliable) to give an idea of what we could do

PR #370, opened 2025-05-18
fix(utils): Removed forced path for mksquashfs

RezaRahemtola:
Fixes an issue with `aleph program upload` on operating systems where `mksquashfs` is not installed at this location (eg on MacOS with homebrew it's at `/opt/homebrew/bin/mksquashfs`, and `/usr/bin` is not writable without disabling important security features of the OS).

Letting the OS figures out the location of the binary to use fixes this & works for everybody.

## Print screen / video

Before:
image

After:
image


RezaRahemtola:
@nesitor small bump to merge this ;)

nesitor:
> @nesitor small bump to merge this ;)

@RezaRahemtola the PR is not passing the CI tests and either the code quality rules. Code quality can be solved easily with the command `hatch run linting:fmt`, but the tests will require a revision.

PR #381, opened 2025-06-26
Relax CRN version check for Confidentials and PAYG

nesitor:
Fix: Relax CRN version check to use custom release also for confidentials and PAYG.


1yam (approved):
LGTM

PR #415, opened 2025-11-07
fix(utils): Removed forced path for mksquashfs

RezaRahemtola:
PR recreated for #370

Enhanced it to find the path of `mksquashfs` to solve the initial issue + pass the linting


RezaRahemtola:
cc @nesitor, closed the previous PR (#370) by mistake

PR #418, opened 2025-11-21
Fix Holding Tier restrictions

nesitor:
Fix: Removed tier 4 from the allowed tier list for holding.

- [X] An LLM was used to review the new code and look for simplifications.
- [X] New classes and functions contain docstrings explaining what they provide.
- [X] All new code is covered by relevant tests.

## Changes

Removed tier 4 from that allowed tier list for holding tier

## How to test

Try to create a big tier VM (more than 4 compute units) and the CLI will not allow it.


1yam (approved):
LGTM

PR #436, opened 2026-02-10
Feat: Add website cmds

aliel:
Explain what problem this PR is resolving

Related ClickUp, GitHub or Jira tickets : ALEPH-XXX

- [ ] The new code clear, easy to read and well commented.
- [ ] New code does not duplicate the functions of builtin or popular libraries.
- [ ] An LLM was used to review the new code and look for simplifications.
- [ ] New classes and functions contain docstrings explaining what they provide.
- [ ] All new code is covered by relevant tests.

## Changes
- [x] Add `aleph website create`
- [x] Add `aleph website update`
- [x] Add `aleph website delete`
- [x] Add `aleph website list`
- [x] Add `aleph website history` (display, prune, restore)
- [ ] Add all tests
- [ ] Check custom domain commands

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,8 @@ dependencies = [
   "ledgerblue>=0.1.48",
   "ledgereth>=0.10",
   "py-sr25519-bindings==0.2",    # Needed for DOT signatures
+  "py-multibase>=1,<2",          # Needed for IPFS CID v0 to v1 conversion
+  "py-multicodec<0.3",           # Needed for IPFS CID v0 to v1 conversion
   "pydantic>=2",
   "pygments==2.19.1",
   "pynacl==1.5",                 # Needed now as default with _load_account changement
--- a/src/aleph_client/__main__.py
+++ b/src/aleph_client/__main__.py
@@ -15,6 +15,7 @@ from aleph_client.commands import (
     node,
     pricing,
     program,
+    website,
 )
 from aleph_client.utils import AsyncTyper
 
@@ -35,6 +36,7 @@ app.add_typer(credit.app, name="credits", help="Credits commands oAleph Cloud")
 app.add_typer(domain.app, name="domain", help="Manage custom domain (DNS) on Aleph Cloud")
 app.add_typer(node.app, name="node", help="Get node info on Aleph Cloud")
 app.add_typer(about.app, name="about", help="Display the information of Aleph CLI")
+app.add_typer(website.app, name="website", help="Manage IPFS-hosted websites on Aleph Cloud")
 
 app.command("pricing")(pricing.prices_for_service)
 
--- a/src/aleph_client/commands/help_strings.py
+++ b/src/aleph_client/commands/help_strings.py
@@ -82,3 +82,9 @@ PORT_FORWARDER_PROTOCOL = "Protocol to forward (tcp, udp, or both)"
 PORT_FORWARDER_TCP = "Enable TCP forwarding for this port"
 PORT_FORWARDER_UDP = "Enable UDP forwarding for this port"
 STORAGE_ENGINE = "Storage engine to use. Either IPFS or Aleph Storage"
+
+# Website help strings
+WEBSITE_NAME = "Unique name for your website"
+WEBSITE_PATH = "Path to your static website directory (e.g., build output like `out` or `dist`)"
+WEBSITE_CID = "IPFS CID v0 of an already uploaded static website directory"
+WEBSITE_FRAMEWORK = "Framework used to build the website (e.g., next, react, vue)"
--- a/src/aleph_client/commands/utils.py
+++ b/src/aleph_client/commands/utils.py
@@ -42,6 +42,17 @@ from aleph_client.utils import fetch_json
 logger = logging.getLogger(__name__)
 
 
+def ipfs_cid_v0_to_v1(cid_v0: str) -> str:
+    """Convert an IPFS CID v0 (base58, Qm...) to CID v1 (base32)."""
+    import base58
+    import multibase
+    import multicodec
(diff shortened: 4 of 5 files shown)


foxpatch-aleph (changes requested):
The PR adds a solid set of `aleph website` commands with good overall structure (aggregate-based registry, version history, rich tables). However, there are several real bugs: a file handle leak in `_upload_directory`, a history duplication bug in `--restore`, and a name sanitization inconsistency that will cause `update`/`delete`/`history` to silently fail on names that were sanitized at deploy time. There are also no tests at all (explicitly flagged in the checklist) and some minor dead-code/UX issues to address.

**`src/aleph_client/commands/website.py`** (line 68): **Bug – file handle leak**: The `open()` calls happen *before* the `try` block. If any `open()` raises (e.g. a file becomes unreadable mid-iteration), all previously opened handles are leaked because the `finally` block is never reached.

Move the entire loop inside the `try`, or use `contextlib.ExitStack` to manage the handles:
```python
import contextlib

async def _upload_directory(directory: Path) -> Optional[str]:
    params = {"recursive": "true", "wrap-with-directory": "true"}
    with contextlib.ExitStack() as stack:
        files = {
            str(file_path.relative_to(directory)): stack.enter_context(open(file_path, "rb"))
            for file_path in directory.rglob("*")
            if file_path.is_file()
        }
        # ... rest of the upload logic
```

**`src/aleph_client/commands/website.py`** (line 76): **Potential bug – aiohttp multipart format for IPFS**: The IPFS `/api/v0/add` endpoint with `recursive=true` expects `multipart/form-data` where each part's `Content-Disposition` carries the file's relative path in the `filename` attribute. When you pass a plain `dict` of file objects to `data=`, aiohttp sets the dict key as the `name` attribute, not `filename`. IPFS uses `filename` to reconstruct the directory tree, so the upload will likely produce a flat structure or fail to preserve paths.

Use `aiohttp.FormData` explicitly:
```python
form = aiohttp.FormData()
for rel_path, fh in files.items():
    form.add_field("file", fh, filename=rel_path)
response = await session.post(url, params=params, data=form)
```

**`src/aleph_client/commands/website.py`** (line 185): **Bug – name sanitization inconsistency**: `deploy` silently sanitizes the name (e.g. `"My Website"` → `"my-website"`), but `update`, `delete`, and `history` do not. A user who deployed `"My Website"` will get `"Website 'My Website' not found"` when trying to `update` or `delete` it, with no hint about why.

Either apply `_sanitize_name` in all commands that accept a name, or—better—show a warning when sanitization changes the input:
```python
sanitized = _sanitize_name(name)
if sanitized != name:
    typer.echo(f"Note: name sanitized to '{sanitized}'")
name = sanitized
```

**`src/aleph_client/commands/website.py`** (line 389): **Bug – restored version remains in history**: After restoring version N, the code saves `site_history[str(current_version)] = current_volume_id` (correctly archiving the current version), but the restored version N is still in `site_history` pointing to the same `volume_id` now live as the new current. This creates a duplicate: version N appears both as the active version and as a history entry.

Remove the restored version from history before saving:
```python
restored_volume_id = site_history.pop(version_key)  # removes it from history
current_volume_id = site["volume_id"]
current_version = site.get("version", 1)
site_history[str(current_version)] = current_volume_id  # archive current
```

**`src/aleph_client/commands/website.py`** (line 337): **Missing confirmation on destructive operation**: `delete` permanently removes the website registry entry with no confirmation prompt. Every other CLI command of this class (e.g. `instance delete`) asks for `--yes` or prompts the user. Consider adding:
```python
if not typer.confirm(f"Delete website '{name}'? This cannot be undone."):
    raise typer.Abort()
```

**`src/aleph_client/commands/website.py`** (line 118): **Unnecessary account load for read-only command**: `list_websites` calls `load_account(...)` even when `--address` is explicitly provided, which may fail if no private key is configured. Since `AlephHttpClient` (used below) is unauthenticated, the account is only needed as an address fallback. Consider:
```python
if not address:
    address = settings.ADDRESS_TO_USE
    if not address:
        account = load_account(...)
        address = account.get_address()
```
Same issue in the display-only branch of `history` (line 433).

**`src/aleph_client/commands/utils.py`** (line 45): **Dead code**: `ipfs_cid_v0_to_v1` is defined here but never imported or called anywhere in the codebase. The `WEBSITE_CID` help string and this function suggest a `--cid` shortcut option for `deploy` was planned but not implemented. Either implement the feature or remove the function (and the two new dependencies in `pyproject.toml`).

Additionally, the CID conversion logic itself appears incorrect: a CIDv1 is structured as `<version-varint><codec-varint><multihash>`, but the code calls `multicodec.add_prefix('cidv1', decoded)` which only prepends the cidv1 multicodec tag to the raw CIDv0 multihash bytes—it never inserts the dag-pb codec varint (`0x70`). This would produce an invalid CID.

**`src/aleph_client/commands/website.py`** (line 46): **Silent exception swallowing**: `_fetch_websites_aggregate` catches all exceptions and returns `{}`. This means a network error, auth failure, or API outage will silently look like "no websites found", leading to confusing errors (e.g. `deploy` will succeed even if the duplicate-check fetched `{}` due to a transient failure). Consider at minimum logging the exception at `DEBUG` level, and re-raising non-`404` HTTP errors.

**`src/aleph_client/commands/website.py`** (line 1): **No tests**: The checklist explicitly marks tests as incomplete. The `_sanitize_name`, `_get_domains_for_volume`, `_upload_directory`, and all five command handlers should have unit/integration tests before merge, following the pattern of `tests/unit/test_*.py` elsewhere in the project. At minimum, `_sanitize_name` edge cases (empty string, all-special-chars, unicode) and the history restore/prune logic should be covered.

PR #437, opened 2026-02-21
Add new methods to manage instance server (reinstall, backup create, info, download, delete 

aliel:
related to:
https://github.com/aleph-im/aleph-sdk-python/pull/279
https://github.com/aleph-im/aleph-vm/pull/874


src/aleph_client/commands/instance/backup.py
```diff
@@ -0,0 +1,457 @@
...
+        if status == 202 and background:
+            with console.status("Backup in progress..."):
+                while status == 202:
+                    await asyncio.sleep(5)
+                    status, result = await manager.create_backup(
+                        vm_id=vm_id,
+                        include_volumes=include_volumes,
+                        skip_fsfreeze=skip_fsfreeze,
+                        run_async=background,
+                    )
```
odesenfans: Don't you risk looping forever? Why not call the corresponding GET method here instead?

odesenfans: Also, no timeout.

src/aleph_client/commands/instance/backup.py
```diff
@@ -0,0 +1,457 @@
...
+        typer.Option(
+            "--skip-fsfreeze",
+            help="Skip filesystem freeze (no consistency guarantee)",
+        ),
+    ] = False,
+    domain: Annotated[
+        Optional[str],
+        typer.Option(help="CRN domain on which the VM is running"),
+    ] = None,
+    chain: Annotated[Optional[Chain], typer.Option(help=help_strings.PAYMENT_CHAIN_USED)] = None,
```
odesenfans, line 109: Missing `metavar=metavar_valid_chains` on all appearances of `chain` in this file.

src/aleph_client/commands/instance/backup.py
```diff
@@ -0,0 +1,457 @@
...
+    domain = (
+        (domain and sanitize_url(domain))
+        or await find_crn_of_vm(vm_id)
+        or typer.prompt("URL of the CRN (Compute node) on which the VM is running")
+    )
+
+    account: AccountTypes = load_account(private_key, private_key_file, chain=chain)
+
+    async with VmClient(account, domain) as manager:
+        status, result = await manager.delete_backup(vm_id=vm_id, backup_id=backup_id)
```
odesenfans, line 356: No user confirmation required for this one?

odesenfans (changes requested):

foxpatch-aleph (changes requested):
The PR adds useful backup/restore/reinstall commands with good UX (progress bars, warnings, confirmation prompts). However, there is one functional bug that makes the `--background` polling loop create new backups on every poll iteration instead of checking status, and a subtle error-handling bug in `reinstall` where `return 1` silently succeeds from Typer's perspective. There are also no tests for any of the new functionality.

**`src/aleph_client/commands/instance/backup.py`** (line 101): **Bug: polling loop re-invokes `create_backup` instead of checking status.** Each iteration of the `while status == 202` loop calls `manager.create_backup(...)` again, which will attempt to start a *new* backup rather than poll the status of the one already running. This should call a status/get endpoint (e.g. `manager.get_backup(vm_id=vm_id)`) to check whether the in-progress backup has completed.

```python
# Should be something like:
status, result = await manager.get_backup(vm_id=vm_id)
```

**`src/aleph_client/commands/instance/__init__.py`** (line 1170): **Bug: `return 1` does not set the process exit code in a Typer command.** Typer ignores integer return values from command functions; `return 1` here exits the function normally with exit code 0. Every other error path in this PR uses `raise typer.Exit(1)` — use that here for consistency and correctness:

```python
raise typer.Exit(1)
```

**`src/aleph_client/commands/instance/backup.py`** (line 209): **Partial output file left on disk after a failed/interrupted download.** If the connection drops mid-transfer, `output` will contain a partial file with no indication to the user. Consider deleting the output file in an error handler, or writing to a temp file and renaming on success:

```python
tmp = output.with_suffix(output.suffix + '.tmp')
try:
    with open(tmp, 'wb') as f:
        ...
    tmp.rename(output)
except:
    tmp.unlink(missing_ok=True)
    raise
```

**`src/aleph_client/commands/instance/backup.py`** (line 293): **Missing confirmation prompt before destructive `delete` operation.** All other destructive commands (`reinstall`, `restore`) show a warning and ask for confirmation. `delete` silently deletes the backup — a safety net the user may need — without any prompt. Consider adding a `yes_no_input("Delete backup {backup_id}?", default=False)` guard.

**`src/aleph_client/commands/instance/backup.py`** (line 404): **`ProgressFile` class defined inside a `with` block inside a function.** This is hard to read, impossible to unit-test independently, and triggers linter warnings. Extract it as a module-level private class `_ProgressFile`.

**`src/aleph_client/commands/instance/__init__.py`** (line 1161): Minor inconsistency: `reinstall` uses `Prompt.ask(...)` (rich) while all `backup.py` commands use `typer.prompt(...)` for the same CRN URL prompt. Pick one; `typer.prompt` is used everywhere else in the codebase.

**`src/aleph_client/commands/instance/backup.py`** (line 1): **No tests added.** None of the five new commands (`create`, `info`, `download`, `delete`, `restore`) or `reinstall` have test coverage. At minimum, unit tests for the argument-validation logic (mutual-exclusion checks, file-not-found, domain resolution fallback) and mocked happy-path tests for each command would significantly reduce regression risk for these destructive operations.

foxpatch-aleph (changes requested):
The backup/restore feature is well-structured overall — good use of progress bars, clear warning messages, and consistent CLI patterns. However, there are three issues worth blocking on: (1) every `json.loads(result)` call is unguarded and will crash with a raw traceback if the CRN ever returns non-JSON (error HTML, empty body on timeout, etc.); (2) the `restore` command bypasses the SDK entirely by accessing `manager.session` directly, tightly coupling the CLI to VmClient internals; and (3) there are no tests for any of the new commands. Additionally, `ProgressFile` is defined inside the function body and `reinstall` uses `return 1` instead of `raise typer.Exit(1)`, which does not actually set the process exit code.

**`src/aleph_client/commands/instance/backup.py`** (line 109): **Bug: unguarded `json.loads` will crash on non-JSON responses.** If the CRN returns an HTML error page, a plain-text message, or an empty body (e.g. on a timeout), this raises `json.JSONDecodeError` with a raw traceback instead of a clean error. Same issue exists at lines 167, 279, and 450.

```python
try:
    backup_info = json.loads(result)
except json.JSONDecodeError:
    echo(f"Unexpected response from CRN: {result}")
    raise typer.Exit(1)
```

**`src/aleph_client/commands/instance/backup.py`** (line 437): **Design issue: accessing `manager.session` directly.** `VmClient.session` is an internal implementation detail of the SDK. This couples the CLI to the SDK's private HTTP plumbing and will silently break if the SDK renames or restructures the attribute (e.g. moves to `httpx`). The related SDK PR (#279) should expose a dedicated `restore_instance(vm_id, file_like)` method instead. If that method isn't available yet, this should at minimum be flagged as a known technical debt with a comment.

**`src/aleph_client/commands/instance/backup.py`** (line 401): **Style: `ProgressFile` should be a module-level class.** Defining a class inside an async function inside a `with` block is unusual and makes the code harder to read, test, and reuse. Move `ProgressFile` to the top of the module.

**`src/aleph_client/commands/instance/backup.py`** (line 101): **Potential hang: no timeout on the `--follow` polling loop.** If the CRN gets stuck (backup process dies but still returns 202 forever), this loop runs indefinitely. Consider adding a maximum poll count or a wall-clock timeout.

```python
max_polls = 120  # 10 minutes at 5s intervals
for _ in range(max_polls):
    await asyncio.sleep(5)
    status, result = await manager.get_backup(vm_id=vm_id)
    if status != 202:
        break
else:
    echo("Timed out waiting for backup to complete.")
    raise typer.Exit(1)
```

**`src/aleph_client/commands/instance/backup.py`** (line 206): **Minor: partial output file is left on disk if the download fails mid-stream.** The `aiohttp.ClientError` catch handles connection failures before the file is opened, but a mid-stream failure (connection drop after headers) leaves a truncated file at `output` with no warning. Consider writing to a `.tmp` file and renaming on success, or at least printing a note that the file is incomplete.

**`src/aleph_client/commands/instance/backup.py`** (line 290): **UX: `delete` has no confirmation prompt.** `restore` and `reinstall` both show a warning and ask `Proceed? [y/n]`. Deleting a backup is also irreversible — add a similar prompt for consistency.

**`src/aleph_client/commands/instance/__init__.py`** (line 1165): **Bug: `return 1` does not set the process exit code.** In an async Typer command, returning a value has no effect on the exit code. Use `raise typer.Exit(1)` to actually signal failure to the caller, consistent with all the backup commands.

```python
# before
return 1
# after
raise typer.Exit(1)
```

**`src/aleph_client/commands/instance/backup.py`** (line 1): **Missing tests.** The PR adds six new CLI commands with non-trivial logic (polling loops, file upload, progress tracking, two-argument mutex validation). There are no tests for any of them. The existing test suite in `tests/unit/test_instance.py` has a clear pattern for mocking `VmClient` — please add at minimum happy-path tests for `create`, `info`, `download` (both via URL and `--vm-id`), `delete`, and `restore`, and a test for the mutex validation in `download` (both url + vm_id given, neither given).

**`src/aleph_client/commands/instance/backup.py`** (line 450): **Same unguarded `json.loads` as line 109.** The `restore` command parses the 200 response body without catching `JSONDecodeError`. If the CRN returns `"OK"` or any non-JSON 200, this crashes.

foxpatch-aleph (changes requested):
The PR adds a well-structured set of backup/restore/reinstall commands with good UX (warnings, confirmations, progress bars, tmp-file cleanup). However, there are two genuine bugs: the `--follow` polling loop in `create` has no timeout and will hang forever if the CRN gets stuck, and multiple `json.loads(result)` calls are unguarded — any non-JSON 200 response (HTML error page, plain-text message) will produce an unhandled traceback instead of a clean error. There are also no tests for the new module.

**`src/aleph_client/commands/instance/backup.py`** (line 135): **Bug: unbounded polling loop.** `while status == 202` will loop forever if the CRN crashes or never transitions out of the in-progress state. Add a maximum wait time or retry count, e.g.:

```python
max_wait = 600  # seconds
elapsed = 0
while status == 202 and elapsed < max_wait:
    await asyncio.sleep(5)
    elapsed += 5
    status, result = await manager.get_backup(vm_id=vm_id)
if status == 202:
    echo(f"Backup still in progress after {max_wait}s. Check later with: aleph instance backup info {vm_id}")
    raise typer.Exit(1)
```

**`src/aleph_client/commands/instance/backup.py`** (line 143): **Bug: unguarded `json.loads`.** If the CRN returns a 200 with a non-JSON body (e.g., a proxy HTML error page), this raises an unhandled `json.JSONDecodeError`. Same issue exists at lines 196, 306, and 444. Wrap with a try/except:

```python
try:
    backup_info = json.loads(result)
except json.JSONDecodeError:
    echo(f"Unexpected response from CRN: {result}")
    raise typer.Exit(1)
```

**`src/aleph_client/commands/instance/backup.py`** (line 314): **Minor: type narrowing.** At this point `url` is still typed `Optional[str]`, so mypy will flag `_download_from_url(url, output)`. All `None` paths do exit early, so the logic is correct — but an `assert url is not None` (or a cast) makes the intent explicit and satisfies type checkers.

**`src/aleph_client/commands/instance/backup.py`** (line 317): **Usability: no `--yes` / `-y` flag.** `delete` (and `restore`, `reinstall`) call `yes_no_input()` unconditionally, which blocks scripting/automation. A `--yes` flag to skip the prompt is a common CLI convention and is already present in other commands in this codebase.

**`src/aleph_client/commands/instance/backup.py`** (line 1): **Testing: no test coverage.** None of the new commands (`create`, `info`, `download`, `delete`, `restore`) or `reinstall` have tests. At minimum, unit tests mocking `VmClient` for the happy path and error-status branches would be valuable given the complexity of the state machine in `create --follow`.

**`src/aleph_client/commands/instance/backup.py`** (line 431): **Minor: bypassing `VmClient` abstraction.** `manager.session.post(url, headers=header, data=data)` accesses the internal session directly. If `VmClient` ever adds middleware (retries, auth refresh), this path won't benefit. Consider adding a `restore_from_file` method to `VmClient` (or the SDK) to keep the boundary clean — though this may be intentional pending the SDK PR #279.

foxpatch-aleph (changes requested):
This PR adds useful backup and restore functionality for VM instances. However, there are several issues that need to be addressed: (1) The `_download_from_url` function doesn't handle the case when `response.content_length` is `None`, which will cause the Progress bar to fail; (2) The `_ProgressFile` class inherits from `io.RawIOBase` but doesn't properly implement the required interface; (3) The `restore` command directly accesses `manager.session` which may not be initialized in all cases; (4) The `reinstall` command's logic for `erase_volumes` is confusing (double negative). These issues should be fixed before merging.

**`src/aleph_client/commands/instance/backup.py`** (line 228): When `response.content_length` is `None` (common with some servers), the Progress bar will fail. Handle this case:

```python
total = response.content_length
if total is None:
    task = progress.add_task(f"Downloading to {output}", total=-1)  # -1 means unknown total
else:
    task = progress.add_task(f"Downloading to {output}", total=total)
```

**`src/aleph_client/commands/instance/backup.py`** (line 36): The `_ProgressFile` class inherits from `io.RawIOBase` but doesn't implement all required abstract methods. This could cause issues. Consider either:
1. Implementing the full interface (add `writable()`, `seekable()`, `seek()`, `tell()`, etc.)
2. Or simply passing the file object directly and using a callback for progress updates instead.

**`src/aleph_client/commands/instance/backup.py`** (line 431): Directly accessing `manager.session` bypasses the context manager pattern. The session should be accessed through the proper API. Consider checking if the session exists and handling the case where it might not be available.

**`src/aleph_client/commands/instance/__init__.py`** (line 1167): The logic `erase_volumes=not keep_data` is a double negative that's confusing. Consider renaming the parameter for clarity, e.g., `keep_data` instead of `erase_volumes`, or add a comment explaining the inversion.

**`src/aleph_client/commands/instance/backup.py`** (line 237): The file write operation inside the async loop should ensure proper flushing. Consider adding `f.flush()` periodically or using `os.fsync()` to prevent data loss on crash during download.

**`src/aleph_client/commands/instance/backup.py`** (line 307): Variable `url` is reassigned here, shadowing the function parameter. Consider renaming to `download_url` for clarity.

**`src/aleph_client/commands/instance/backup.py`** (line 410): The `get_restore_endpoint()` call returns `(url, header)` but there's no error handling. If the endpoint is unavailable, this should be caught and reported to the user.

**`src/aleph_client/commands/instance/__init__.py`** (line 1158): The domain resolution logic here duplicates the `_resolve_domain` helper from `backup.py`. Consider refactoring to use a shared helper function to avoid code duplication.

**`.github/workflows/pytest.yml`** (line 48): The constraint `virtualenv<21` is a workaround for a compatibility issue. Consider adding a comment explaining why this constraint is needed and when it might be removable.

PR #444, opened 2026-04-11
fix(cli): rename item_hash to vm_id in instance commands for consistency

aliel:


foxpatch-aleph (approved):
This PR correctly renames the CLI parameter `item_hash` to `vm_id` in instance and port forwarder commands for consistency. The internal API calls properly continue to use `item_hash` as the parameter name (the API contract), while the CLI-facing parameter is now more user-friendly. All tests have been updated accordingly. The test improvement in `test_no_format.py` to strip ANSI codes is a nice bonus that makes tests more robust.

**`src/aleph_client/commands/instance/__init__.py`** (line 910): Minor suggestion: The help text "VM item hash to delete" could be simplified to "VM ID to delete" for consistency with the parameter name.

**`src/aleph_client/commands/instance/port_forwarder.py`** (line 38): The docstring for `list_ports` still says "specific item hash" - consider updating to "specific VM" for consistency.

**`tests/unit/test_no_format.py`** (line 61): Nice improvement - stripping ANSI codes makes the test more robust against terminal formatting changes.

PR #445, opened 2026-04-15
fix(cli): add --metadata to file upload/pin and fix argument consistency

aliel:


foxpatch-aleph (changes requested):
The PR adds a useful `--metadata` option to `pin` and `upload` commands and fixes argument consistency issues. However, there is a resource leak bug in the `upload_directory()` function where file handles are opened but never closed. Additionally, the new `--metadata` feature lacks test coverage. These issues should be addressed before merging.

**`src/aleph_client/commands/files.py`** (line 204): **Bug: Resource leak** - File handles are opened in the `files` dictionary but never closed. This can lead to file descriptor exhaustion when uploading large directories. The files should be closed after the IPFS upload completes, or use a context manager pattern.

**`src/aleph_client/commands/files.py`** (line 233): **Code quality**: The call to `pin()` uses positional arguments which is fragile. Consider using keyword arguments for clarity and to prevent breakage if the function signature changes: `await pin(item_hash=cid, channel=channel, ..., metadata=metadata, debug=debug)`

**`tests/unit/test_commands.py`** (line 747): **Missing tests**: The new `--metadata` option added to `pin` and `upload` commands has no test coverage. Add tests for: (1) valid metadata parsing with `--metadata key=value`, (2) invalid metadata format handling (missing `=`), (3) multiple metadata entries, (4) metadata being passed correctly to `create_store()` via `extra_fields`.

**`src/aleph_client/commands/files.py`** (line 46): **Minor**: Consider using `typer.Abort()` instead of `typer.echo()` + `typer.Exit(code=1)` for cleaner error handling, or raise a custom exception.

PR #447, opened 2026-04-20
feat(cli): pre-check CRN capacity via reserve_resources before create

aliel:


foxpatch-aleph (commented):
The capacity pre-check feature is correctly implemented and follows the existing codebase patterns well. The logic is sound, the error handling covers the main cases (503 = no capacity, 400 = bad request, other = warning), and the placement before payment/publishing is appropriate. However, the existing tests do not verify that reserve_resources is actually called, and there are no tests covering the failure scenarios (503, 400). Adding assertions to the existing parametrized tests and adding dedicated failure-scenario tests would provide proper confidence in the feature.

**`src/aleph_client/commands/instance/__init__.py`** (line 748): Variable name `status` shadows the `status` variable used later at line 763 for `message, status = await client.create_instance(...)`. While Python scoping makes this safe, renaming one of them (e.g. `precheck_status` or `crn_status`) would improve readability.

**`tests/unit/test_instance.py`** (line 547): Add `mock_vm_client.reserve_resources.assert_called_once()` to the test cases where `payment_type` is superfluid/credit or the instance has confidential/gpu. This would verify the pre-check is actually invoked in the happy path.

**`tests/unit/test_instance.py`** (line 359): Add test cases for failure scenarios: (1) reserve_resources returns 503 — verify typer.Exit with code=1 without calling create_instance, (2) reserve_resources returns 400 — verify typer.Exit with code=1, (3) reserve_resources returns a non-200/400/503 status — verify the warning is emitted and the flow proceeds to create_instance.

foxpatch-aleph (approved):
This PR adds a capacity pre-check before instance creation by calling the CRN's `reserve_resources` endpoint as a dry-run admission control step. The implementation correctly gates the pre-check behind the same conditions that require a CRN (stream/confidential/gpu/credit), properly handles the three response categories (503/400 abort with clear messages, other non-200 warns but proceeds), and is fully covered by new parametrized tests plus an assertion added to the existing test. The diff is focused, well-structured, and follows the existing patterns of the codebase.

**`src/aleph_client/commands/instance/__init__.py`** (line 759): Nit: the two adjacent f-strings (`f"..." f"..."`) work due to Python's implicit string concatenation but look unusual. This could be a single f-string: `echo(f"[yellow]Warning:[/yellow] capacity pre-check returned {precheck_status}: {precheck_result}")`

foxpatch-aleph (commented):
The PR adds a capacity pre-check before instance creation by calling reserve_resources on the selected CRN. The logic is correct, tests are well-structured, and the pre-check happens before any financial commitment (flow creation). The main observations are: (1) reserve_resources depends on a cross-repo SDK change not yet visible in the installed dependencies, (2) the data shape sent to reserve_resources (model_dump JSON) differs from what's sent to create_instance (raw dict), which is a potential mismatch if the CRN endpoint expects the same format as the create call, and (3) exception handling for network errors during the pre-check follows existing patterns (no try/except) which could produce confusing tracebacks.

**`src/aleph_client/commands/instance/__init__.py`** (line 748): The `reserve_resources` method does not exist on `VmClient` in the currently installed `aleph-sdk-python` version. This PR depends on a companion SDK change — ensure that change is released or co-merged.

**`src/aleph_client/commands/instance/__init__.py`** (line 748): `content.model_dump(mode="json", exclude_none=True)` serializes types (Decimal→str, Enum→str) and drops None fields, which produces a different data shape than the `**content_dict` passed to `create_instance`. Ensure the CRN's `reserve_resources` endpoint expects this serialized form.

**`src/aleph_client/commands/instance/__init__.py`** (line 746): If `reserve_resources` raises a network error (e.g. connection refused from unreachable CRN), there is no try/except around the VmClient context manager. The exception will propagate as an unhandled traceback. Consider wrapping in a try/except to produce a user-friendly error message, consistent with the cost estimation block above.

foxpatch-aleph (approved):
Well-structured PR that adds a CRN capacity pre-check before instance creation for stream/confidential/gpu/credit instances. The logic correctly fails early on 503 (no capacity) and 400 (rejected request), warns on other non-200 statuses, and is placed after cost estimation but before payment/publishing. Tests cover all three branches with clear parametrization. The only edge case is precheck_status=None (network error) which falls through with a warning — acceptable behavior, though an explicit None check would be slightly clearer.

**`src/aleph_client/commands/instance/__init__.py`** (line 748): precheck_status can be None (returned by VmClient.reserve_resources on ClientError). The current code treats this the same as other non-200 responses (warning + proceed). Consider an explicit `if precheck_status is None` branch for clarity.

PR #450, opened 2026-04-27
fix(cli): fall back when file list endpoint returns 404

NiKrause:
## Summary
Add a fallback for `aleph file list` when `/api/v0/addresses/{address}/files` returns 404.

## Changes
- try the legacy files endpoint first
- fall back to querying STORE messages via `messages.json`
- keep non-404 errors surfaced directly
- add focused tests for the fallback path

## Context
This started failing for us around April 26, 2026 at 1:00 PM, while balance and messages endpoints still returned 200.

### DIFF
--- a/src/aleph_client/commands/files.py
+++ b/src/aleph_client/commands/files.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import json as json_lib
 import logging
 import time
-from datetime import datetime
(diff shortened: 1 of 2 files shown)


NiKrause:
The old `/api/v0/addresses/{address}/files` endpoint appears to be working again now. This issue looked temporary on the backend side.

PR #451, opened 2026-05-26
fix(files): preserve directory tree on `file upload`

aliel:
Directory uploads sent each file as a dict entry, which aiohttp encoded
  with the relative path as the multipart field name. Kubo's /api/v0/add
  requires every part named "file" with the path in the filename, so
  subdirectories and nested files were dropped.

  Build the multipart with aiohttp.FormData: name every part "file", put
  the relative path in the filename, emit application/x-directory parts for
  directories, and close file handles after upload.

### DIFF
--- a/src/aleph_client/commands/files.py
+++ b/src/aleph_client/commands/files.py
@@ -250,7 +250,7 @@ async def upload(
             cid = await upload_directory(path)
             if not cid:
                 typer.echo("CID not found in response.")
-                typer.Exit(code=1)
+                raise typer.Exit(code=1)
             await pin(cid, channel, private_key, private_key_file, chain, ref, address, metadata, debug)
         else:
             typer.echo(f"Error: File not found: '{path}'")
--- a/src/aleph_client/commands/files.py
+++ b/src/aleph_client/commands/files.py
@@ -195,30 +195,51 @@ async def upload(
             typer.echo(f"Upload directory {path}...")
 
             async def upload_directory(directory: Path):
+                # Kubo's /api/v0/add expects each entry as a multipart part named "file";
+                # the filename (with slashes) encodes the relative path, so the directory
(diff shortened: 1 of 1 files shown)


foxpatch-aleph (changes requested):
The PR correctly fixes directory uploads by switching from a dict-based multipart to proper aiohttp.FormData with Kubo-compatible parts. However, there is a genuine bug on line 253 where `typer.Exit(code=1)` is missing the `raise` keyword, causing the error case to silently fall through and call `pin(None, ...)`. The file handle management in the `finally` block is acceptable but could be more robust.

**`src/aleph_client/commands/files.py`** (line 253): Missing `raise` keyword: `typer.Exit(code=1)` should be `raise typer.Exit(code=1)`. Without this, when no CID is returned, the code silently falls through and calls `pin(None, channel, ...)` which will fail in an obscure way.

foxpatch-aleph (approved):
This PR correctly fixes directory uploads to Kubo's IPFS API by switching from a dict-based multipart (where aiohttp encodes dict keys as field names) to a FormData-based approach where every part is named "file" and the relative path goes in the filename. This is exactly what Kubo's /api/v0/add requires for preserving directory trees. The changes are correct and well-structured: handles are safely closed in a finally block, files are sorted for deterministic ordering, and there's an important latent bug fix where typer.Exit is now properly raised instead of silently discarded. The only minor concern is a potential file handle leak if open() fails mid-loop, as the try/finally scope doesn't cover file opening.

**`src/aleph_client/commands/files.py`** (line 214): If an exception occurs during open() (e.g. permissions issue on the 5th file), the handles opened so far would leak since file closing is inside the try/finally block that starts at line 225. Consider wrapping the file-opening loop in the try block, or collecting file paths first and opening inside the try.

PR #452, opened 2026-05-28
Fix sdk version

aliel:


foxpatch-aleph (approved):
This PR adjusts the aleph-sdk-python dependency from a strict pinned version (2.3) to a compatible patch range (2.3.*) and adds the [dns] extra. This is a reasonable change: pinning to ==2.3 was overly restrictive since patch releases should be backward compatible, and the [dns] extra likely includes required DNS resolution dependencies. The comment alignment whitespace fixes are cosmetic and harmless. No functional, security, or correctness issues.

PR #454, opened 2026-05-28
Fix sdk version

aliel:


foxpatch-aleph (approved):
Pins aleph-sdk-python[dns] from a wildcard constraint (2.3.*) to a specific patch version (2.3.4), improving dependency reproducibility without changing the allowable version range.

foxpatch-aleph (approved):
Correctly adds the `[dns]` extra to the `aleph-sdk-python` dependency, which pulls in `tldextract` — a transitive dependency required by `DomainValidator` from `aleph.sdk.domain`. The change is syntactically valid and the version pin is preserved. There's a trivial whitespace-only change in a comment that is harmless.

**`pyproject.toml`** (line 33): Cosmetic: trailing whitespace added before the inline comment. This is non-blocking but inconsistent with the other comments in the file.

